PR #956, opened 2016-05-24
fix(customFacets): 2467 change to POST

cy:
- no longer caches because POSTs don't cache
- fixes filter in IE not working
- show error toast when POST failed
Closes #2467


cy:
replaced by https://github.com/NCI-GDC/portal-ui/pull/1046

PR #974, opened 2016-05-26
feat(app): 640 expand / collapse facet sidebar

alex-wilmer:
Closes #640

### DIFF
--- a/app/scripts/search/search.controllers.ts
+++ b/app/scripts/search/search.controllers.ts
@@ -45,6 +45,7 @@ module ngApp.search.controllers {
     tabSwitch: boolean = false;
     projectIdChartConfig: any;
     primarySiteChartConfig: any;
+    facetsCollapsed: boolean = false;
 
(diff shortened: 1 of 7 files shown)


app/scripts/search/search.controllers.ts
```diff
@@ -225,6 +226,11 @@ module ngApp.search.controllers {
       this.$state.go("query." + toTab, stateParams, { inherit: true });
     }
 
+    toggleFacets(shouldCollapse) {
```
shanewilson: how come this is only needed in search? not projects/annotations ctrl?

shanewilson: ^ @alex-wilmer

alex-wilmer: Sorry missed this.. it was only needed to send a message to the pie chart directive to refresh itself, and projects/annotations don't have pie charts.

shanewilson:
👍

PR #978, opened 2016-05-30
feat(gql): 2350 find/report bogus fields.

ithinkicancode:
Closes #2350


alex-wilmer:
This seems to work, the code is fine albeit a little confusing in places. I would rather see this tested and merged in before refactoring it. I probably just need to do more purescript while listening to 90's brit pop and it will click faster.

alex-wilmer:
👍

PR #985, opened 2016-06-01
fix(table): truncated cell fit UUID, break longer

alex-wilmer:
TODO:
- [x] better name for class
- [x] headers break word not all
- [x] make separate classes... not all fields that should have similar width should break-all


shanewilson:
👍

shanewilson:
@cy / @ithinkicancode ?

ithinkicancode:
👍

PR #987, opened 2016-06-01
Feat/2230 file page add proj

cy:
there's a blue note on the mockup that says 'add project_id to download file metadata in the cart, do we need a ticket?' doing that here too

todo:
- [x] display in file properties instead


shanewilson:
👍

PR #995, opened 2016-06-03
fix(tree): 2572 expand first entity with children

alex-wilmer:
TODO:
- [x] group entities with fake parents
- [x] add comments to BiospecimenService
- [x] more test cases
- [x] use stitching algo instead of grouping paths

### DIFF
--- a/app/scripts/components/ui/biospecimen/biospecimen.controllers.ts
+++ b/app/scripts/components/ui/biospecimen/biospecimen.controllers.ts
@@ -27,7 +27,7 @@ module ngApp.components.ui.biospecimen.controllers {
         $scope.participant.samples.expanded = true;
         this.activeBioSpecimenDoc = $scope.participant.samples[0];
         this.activeBioSpecimenDocType = "sample";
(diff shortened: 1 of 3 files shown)


shanewilson:
does this still have edge cases?

alex-wilmer:
Hmm yes most likely but not for any foreseeable data model we might run into.. I can imagine crazier bio trees with more gaps between actual nodes and placeholders that would make the UI slightly off, but not broken or unusable. I think the code works for what we have + the tests match that.

shanewilson:
👍

alex-wilmer:
@cy @ithinkicancode ? To test this in the app you can visit http://localhost:3000/cases/001c294c-5ae8-55cd-bf07-88112e9b117b and notice that its possible to expand aliquots whereas on develop they are not visible because they are nested under placeholder nodes which are not visible https://gdc-portal.nci.nih.gov/cases/001c294c-5ae8-55cd-bf07-88112e9b117b

cy:
👍

PR #999, opened 2016-06-06
Feat/1222 id fields in custom facets

cy:
image
forgot about facet-free-text already having a autocomplete false option, reusing that. 

todo:
- [x] remove autocomplete not supported
- [x] remove 🔍
- ~~don't disallow *~~ nvm, * breaks OR
- ~~time travel btn~~ part of new ticket 2578

### DIFF
--- a/app/scripts/components/facets/facets.services.ts
+++ b/app/scripts/components/facets/facets.services.ts
@@ -242,7 +242,6 @@ module ngApp.components.facets.services {
       return _.map(_.filter(data, (datum) => {
         return datum.doc_type === docType &&
                datum.field !== 'archive.revision' &&
-               !_.includes(datum.field, "_id") &&
                !_.includes(current, datum.field) &&
                !_.includes(docType === 'files'
                 ? _.pluck(this.SearchTableFilesModel.facets, "name")
@@ -316,11 +315,19 @@ module ngApp.components.facets.services {
     }
 
     addField(docType: string, fieldName: string, fieldType: string): void {
+      var facetType = 'terms';
+      if (_.includes(fieldName, 'datetime')) {
+        facetType = 'datetime';
+      } else if (_.some(['_id', '_uuid', 'md5sum'], a => _.includes(fieldName, a))) {
+        facetType = 'id';
+      } else if (fieldType === 'long') {
+        facetType = 'range';
+      }
       this.fieldsMap[docType].unshift({
           name: fieldName,
           title: fieldName,
           collapsed: false,
-          facetType: fieldType === 'long' ? 'range' : _.includes(fieldName, 'datetime') ? 'datetime' : 'terms',
+          facetType: facetType,
           removable: true
       });
       this.save();
--- a/app/scripts/components/facets/facets.directives.ts
+++ b/app/scripts/components/facets/facets.directives.ts
@@ -54,7 +54,9 @@ module ngApp.components.facets.directives {
         field: "@",
         entity: "@",
         template: "@",
-        autocomplete: "@"
+        autocomplete: "@",
+        removeFunction: "&",
+        removable: "@",
       },
       replace: true,
       templateUrl: "components/facets/templates/facets-free-text.html",
--- a/app/scripts/components/facets/templates/facets-free-text.html
+++ b/app/scripts/components/facets/templates/facets-free-text.html
@@ -8,7 +8,7 @@
       role="button"
       for="{{ 'autocomplete-' + title }}"
     >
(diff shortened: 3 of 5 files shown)


shanewilson:
did @Fgerthoffert have some suggestions for those id search fields? Like Search for id -> Enter id, remove the icon, get rid of the 'autocomplete not supported' dropdown? Or was that a different ticket?

shanewilson:
> don't disallow * nvm, * breaks OR

what is this?

cy:
screen shot 2016-06-16 at 10 29 24 am
those suggestions are done in https://github.com/NCI-GDC/portal-ui/pull/999/commits/4e1301b534dfa3d8ac61e911eb409ecd4774b981

shanewilson:
👍

cy:
I meant allowing *'s to be typed in the input box. We tried if prefix search works here, and it does so was going to allow *'s to be typed. But having 2 values doesn't work. For example  Demographic Id IS 11* => 61 cases.  Demographic Id IN (11*, 00*) => still 61 cases, they all start with 11.

shanewilson:
all the case ids start with 11?

shanewilson:
the API should handle multiple `*`
https://github.com/NCI-GDC/gdcapi/blob/develop/gdcapi/esutils/request.py#L468

cy:
`demographic.demographic_id`s. `Demographic Id IS 11*` => 61, `Demographic Id IS 00*` => 60. `Demographic Id IN (11*, 00*) ` => 61. Shouldn't it be 61+ 60? The actual prefix facet component hides the input box if one is active. https://github.com/NCI-GDC/portal-ui/blob/develop/app/scripts/components/facets/templates/facets-prefix.html#L28

shanewilson:
not if the cases that have 11 dem id also have 00 dem ids

cy:
but demographic is one-to-one https://github.com/NCI-GDC/gdcdictionary/blob/develop/gdcdictionary/schemas/demographic.yaml#L28 & I was on terry's support missing gdcapi branch. On develop I get 
```
  File "/Users/CYu/repos/portal-api/gdcapi/esutils/request.py", line 123, in _get_term_or_prefix_filter
    if '*' in value[0]:
IndexError: list index out of range```
interesting?

File: app/scripts/components/facets/facets.services.ts [hunk 1]
```diff
@@ -316,11 +315,19 @@ module ngApp.components.facets.services {
     }
 
     addField(docType: string, fieldName: string, fieldType: string): void {
+      var facetType = 'terms';
+      if (_.includes(fieldName, 'datetime')) {
+        facetType = 'datetime';
+      } else if (_.includes(fieldName, '_id') || _.includes(fieldName, '_uuid') || _.includes(fieldName, 'md5sum')) {
```
ithinkicancode: how about `else if (['_id', '_uuid', 'md5sum'].some (a => _.includes (fieldName, a))) {`?

ithinkicancode:
@cy, don't think I changed `_get_term_or_prefix_filter` but I'll have a look.

cy:
@ithinkicancode from http://localhost:5000/cases?filters={"op":"and","content":[{"op":"in","content":{"field":"cases.demographic.demographic_id","value":["11*","00*"]}}]} seeing 
```
body = {'query': {'filtered': {'filter': {'bool': {'must': [{'prefix': {u'demographic.demographic_id': u'11'}}]}}, 'query': {'match_all': {}}}}}
doc_type = case
params = {'sort': '', 'from': 0, '_source': 'submitter_aliquot_ids,sample_ids,portion_ids,updated_datetime,created_datetime,slide_ids,submitter_portion_ids,submitter_analyte_ids,analyte_ids,submitter_id,state,case_id,submitter_sample_ids,days_to_index,aliquot_ids,submitter_slide_ids', 'facets': '', 'filters': u'{"op":"and","content":[{"op":"in","content":{"field":"cases.demographic.demographic_id","value":["11*","00*"]}}]}', 'query': '', 'expand': '', 'size': 10}
index = gdc_from_graph
```
you're right looks like the 2nd prefix term isn't in body

phuongmy:
The feature looks good on Dev. 
However the Case submitter ID prefix search does not work anymore although it's working on Prod. There might be a conflict?

cy:
that took me way too long to figure out, it was so simple. Doctype for prefix search was declared twice after I moved some things around. Thanks for finding it @phuongmy! Deployed to dev again.

phuongmy:
👍

PR #1008, opened 2016-06-10
fix(quicksearch): 2513 unhighlight, open on enter

cy:
- remove ctrl+space listener (didn't work)
Closes #2513

### DIFF
--- a/app/scripts/components/quick-search/quick-search.directives.ts
+++ b/app/scripts/components/quick-search/quick-search.directives.ts
@@ -1,14 +1,5 @@
 module ngApp.components.quickSearch.directives {
-  enum KeyCode {
-    Space = 32,
-    Enter = 13,
-    Esc = 27,
-    Left = 37,
-    Right = 39,
-    Up = 38,
-    Down = 40,
-    Tab = 9
-  }
+  import KeyCode = ngApp.components.facets.controllers.KeyCode;
 
   /* @ngInject */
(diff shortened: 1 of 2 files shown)


alex-wilmer:
👍

app/scripts/components/quick-search/quick-search.directives.ts
```diff
@@ -1,17 +1,11 @@
...
-    Down = 40,
-    Tab = 9
-  }
 
   /* @ngInject */
-  function QuickSearch($uibModal: any, $window: ng.IWindowService, $uibModalStack): ng.IDirective {
+  function QuickSearch($uibModal: any,
+                       $window: ng.IWindowService,
+                       $uibModalStack,
+                       KEY_CODE: Object
```
shanewilson: doesn't this lose autocomplete?

cy: yeah, but the tests pass and KeyCodes are only declared once. Tradeoff?

shanewilson: I think I would rather have duplicate code with typesafety/autocompletion

cy:
So all I did was reset to the commit where importing enums was used, and now Phantom's ok with it and the tests pass. Hooray, no duplicate code and autocompletion.

shanewilson:
👍

PR #1011, opened 2016-06-13
feat(facet): support missing.

ithinkicancode:
Closes #2107

### DIFF
--- a/app/scripts/components/facets/facets.controllers.ts
+++ b/app/scripts/components/facets/facets.controllers.ts
@@ -116,9 +116,7 @@ module ngApp.components.facets.controllers {
 
       this.terms = terms;
       this.actives = this.FacetService.getActives(this.name, terms);
-      // TODO: Currently there is some complication supporting _missing properly thereby we're hiding
-      // _missing in facets. Once we fully support _missing, #reject should be removed.
-      this.inactives = _.reject(_.difference(terms, this.actives), (term) => term.key === '_missing');
+      this.inactives = _.difference(terms, this.actives);
     }
 
     toggle(event: any, property: string) {
--- a/app/scripts/components/facets/templates/facet.html
+++ b/app/scripts/components/facets/templates/facet.html
@@ -23,7 +23,7 @@
   </h4>
 
   <div data-ng-show="!tc.collapsed">
-    <p data-ng-show="(!tc.terms.length && !tc.error) || (tc.terms.length === 1 && tc.terms[0].key === '_missing')">
+    <p data-ng-show="(!tc.terms.length && !tc.error) ">
       No Data for this Field
     </p>
     <p data-ng-show="tc.error">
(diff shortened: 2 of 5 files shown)


cy:
can't break it so easily anymore but get 
screen shot 2016-06-13 at 3 31 22 pm 
when there's supposed to be 27 for example. 

Do you need to send filter `"op": "is"` instead of `"op": "in"` for missing?

app/scripts/components/facets/facets.controllers.ts
```diff
@@ -98,6 +99,10 @@ module ngApp.components.facets.controllers {
       this.FacetService.addTerm(facet, term);
     }
 
+    addMissing(facet: string): void {
+      this.FacetService.addTerm(facet, 'MISSING');
```
cy: `this.FacetService.addTerm(facet, 'MISSING', 'IS');` ?

cy: then it also needs to do 'OR' for if user clicks both 'missing' and 'alive'?

shanewilson: api will be updated to allow `vital_status in [blood, _missing]` as well as `vital_status = blood or vital_status is missing`

ithinkicancode: `vital_status = blood or vital_status is missing` is already supported on the server side. The adv. search generates the correct filter and the api can consume that data structure. This PR adds the support to `missing` in the `in` filter - therefore the `this.FacetService.addTerm` call @cy mentioned is to add `missing` to the `in` array. Does that make sense?

cy: not clear on your sentence, but the line I suggested sends 'op': 'is'. 'or' is needed in that case but the current facets display doesn't support 'or'. @shanewilson's saying to update the api to support 'in' for missing?

app/scripts/components/facets/facets.controllers.ts
```diff
@@ -98,6 +99,10 @@ module ngApp.components.facets.controllers {
       this.FacetService.addTerm(facet, term);
     }
 
+    addMissing(facet: string): void {
```
shanewilson: why do you need an addMissing function? `_missing` should just work like any other term shouldn't it?

ithinkicancode: removed.

app/scripts/components/facets/templates/facet.html
```diff
@@ -70,6 +71,16 @@
           {{ ::term.key | translate }}
         </span>
       </span>
+      <span data-ng-if="term.key === '_missing'"
```
shanewilson: do we need this?

ithinkicancode: Yup, this is how we can show "No Data" instead of "MISSING". We could translate this in the controller but, since this is a presentation, I do it in here even it involves a bit boilerplate.

shanewilson: why not do this on line 71?

ithinkicancode: If we did this on line 71 and removed the markup tailored to 'MISSING', we'd end up with a lot of ternary expressions wherever `term.key` is referenced. For example, `{{ (term.key === '_missing' ? 'No Data' : term.key) | translate }}` or `data-ng-click="tc.add(tc.name, term.key === '_missing' ? 'MISSING' : term.key)"` Seems to be adding a lot of more noise than having a dedicated markup for `_missing`. Thoughts?

shanewilson:
> data-ng-click="tc.add(tc.name, term.key === '_missing' ? 'MISSING' : term.key)"

why do we need this?

Don't these checks just have to be in two places? the facet and the current query?

ithinkicancode:
'MISSING' is for the filter object. Can `_missing` work? I think the API looks for `missing` or 'MISSING`.

shanewilson:
the API looks for `missing` because you set it to look for `missing` - just set it to `_missing` to match the facet response.

https://github.com/NCI-GDC/gdcapi/pull/542/files#diff-d7672a7da14cb1b0e7004bc97d420775R482

ithinkicancode:
there are a couple more places in the api referencing `missing`. Not sure what the impact would be. Do you want me to take that approach? Even without `data-ng-click="tc.add(tc.name, term.key === '_missing' ? 'MISSING' : term.key)"`, we'd still have `data-uib-tooltip="{{ term.key === '_missing' ? 'No Data' : term.key }}"`.

shanewilson:
Your new API code is the only thing that has to change to update `_missing` since this feature didn't exist before. Whether we use two facets or a ternary you should be using the same value for the filter as what comes back in the term facet.

ithinkicancode:
@shanewilson, your concern has been addressed. Now `_missing` is used in both UI and API. You need https://github.com/NCI-GDC/gdcapi/pull/542 for the API.

phuongmy:
Tested UI 1.1.2 @ 60753a5, API 1.1.4 @ a02c407.
First test: select facet gender value = no Data. It seems to work.

But the missing operator on the advanced search got an error.
1. Go to the Advanced Search.
2. Type this query: `cases.demographic.gender IS MISSING`
It throws an exception.

phuongmy:
I just noticed that it does not accept "IS MISSING" in capital letter (I was using the autocomplete while building the query in advanced search).

However, this query works fine: 'cases.demographic.gender is missing'

phuongmy:
Issue 2: The "No Data" is not supported on all the customized facets.
e.g. family history relationship gender does not show "No Data" (but treatment intent works)
screen shot 2016-07-11 at 1 40 23 pm

phuongmy:
Issue 3: the numbers in facet don't add up for clinical filters except demographic.
Same issue for Sample filters. (I did not test the other types of filters)

If you add up the facet values in Gender and Race, the sum gives 15,870 which is the total # cases in the portal.
However, it does not add up for Vital Status, Treatment Therapeutic Agents and Alcohol History for example. There must be an issue with the "No Data" for these filters.
See screenshots below:
1a-vital_status
1b-vital_status
1c-vital_status_issue
2-diagnosis_field
3-exposure_field
4-sample_fields

shanewilson:
missing facets in ES is broken right now

PR #1012, opened 2016-06-13
Feat/2079 tsv

alex-wilmer:
TODO:
- [x] format the fields correctly before sending to special endpoint (send facets along etc)

### DIFF
--- a/app/scripts/components/tables/export-table.service.ts
+++ b/app/scripts/components/tables/export-table.service.ts
@@ -1,0 +1,112 @@
+module ngApp.components.exportTable.services {
+
+  export interface IFilterAndExpands {
+    fields: string[];
+    expand: string[];
+  }
+
+  export interface IExportTableService {
+    getFieldsAndExpand(headings, initial): IFilterAndExpands;
+    exportTable(any): void;
+  }
+
+  class ExportTableService implements IExportTableService {
+    /* @ngInject */
+    constructor(
+      private LocationService,
+      private SearchCasesTableService,
+      private SearchTableFilesModel,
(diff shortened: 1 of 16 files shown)


alex-wilmer:
Projects will not work yet until the endpoint is created on the API. I would prefer this gets looked at / tested before I bother with that.

alex-wilmer:
Added projects support

alex-wilmer:
api link https://github.com/NCI-GDC/gdcapi/pull/546

app/scripts/components/tables/export-table.service.ts
```diff
@@ -0,0 +1,116 @@
...
+      var model = this.getModel(endpoint);
+
+      var tsvHeadings = headings.reduce((acc, x) => ([
+        ...acc,
+        ...(!x.hidden && x.isField
+          ? x.id === `summary.data_categories`
+            ? Object.keys(this.DATA_CATEGORIES).map(k =>
+              `${this.DATA_CATEGORIES[k].abbr}~~~summary.data_categories.${this.DATA_CATEGORIES[k].full}`
+            )
+            : [`${x.name}~~~${x.id}`]
```
shanewilson: why do we need this?

alex-wilmer: I think I can fix this now that I know better how flask handles formdata

alex-wilmer: It was to send an array of key value pairs

alex-wilmer:
@shanewilson @cy feedback addressed.. using proper form data now, but random squigglies for key value pairs

PR #1013, opened 2016-06-13
feat(app): 2573 visited links are darker

cy:
Closes #2573

although visited links become brighter on legacy, I think darker makes more sense?

### DIFF
--- a/app/styles/app.less
+++ b/app/styles/app.less
@@ -297,7 +297,7 @@ a:link {
 }
 
 a:visited {
-  color: #347cb0;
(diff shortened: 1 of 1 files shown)


shanewilson:
yes they should be darker - should probably update legacy too since getting lighter might be confusing.

alex-wilmer:
agreed 👍

cy:
mmkay https://github.com/NCI-GDC/portal-ui-legacy/pull/117 on legacy

shanewilson:
👍

PR #1020, opened 2016-06-22
fix(quickSearch): 617 highlight in pane

cy:
Closes #617 

highlights all fields in the right quicksearch pane that appears in api's 'quick_search_fields'


app/scripts/components/quick-search/templates/participant.html
```diff
@@ -3,11 +3,13 @@
     <i class="fa fa-user fa-12x"></i>
   </div>
 </div>
+<div>
+</div>
```
alex-wilmer: What's this for?

cy: my bad ✂️it

alex-wilmer:
👍

shanewilson:
👍

PR #1023, opened 2016-06-27
fix(facet): hidden checkbox for disabled styles

alex-wilmer:


app/scripts/components/facets/templates/facet.html
```diff
@@ -36,6 +36,7 @@
             tabindex="0"
             class="facet-term-label">
         <i class="fa fa-check-square-o" aria-controls="data-table" aria-checked="false"></i>
+        <input class="hidden" type="checkbox" data-ng-checked="true"  />
```
shanewilson: the aria labels don't match - what happens when you uncheck the checkbox?

shanewilson:
👍

PR #1028, opened 2016-07-04
fix(customFacets): also use actual checkbox

cy:
to be consistent with https://github.com/NCI-GDC/portal-ui/commit/1cdccd50dec123bedc7f8212c2789b8e42d291d6


alex-wilmer:
👍

PR #1032, opened 2016-07-08
feat(gdc-tables): add multi field search

cy:
requires https://github.com/NCI-GDC/gdcapi/pull/555, not sure how to do this requirement 'the Case table should be filtered on the files returned in File table.', also unsure about this 'I should be able to type the exact value ("TCGA-DD-AAVP"), the middle of the value ("DD-A") or the end of the value ("AAVP").'


cy:
🚦 on hold

PR #1044, opened 2016-07-19
refactor(sort): PTRL-116 remove unused save code

cy:
Closes PTRL-116


alex-wilmer:
👍  what was the saved stuff all about?

cy:
it saved the sort settings to local storage and restored the sort at one point. Now it's reading from url only. There's also all this clientSide stuff that I keep deleting and putting back - doesn't look like we're using it anywhere. Scared to touch it though 😰

seancannata:
Regression testing of the sorting by column heads via filters working as expected. See ticket in Jira for more details. PRTL-116

PR #1046, opened 2016-07-22
Fix/prtl 109 custom facets filter

cy:
requires https://github.com/NCI-GDC/gdcapi/pull/568


shanewilson:
did we want to do this? or should it be closed?

cy:
Forgot about this branch completely. I think so - it does fix the 'only facets with values' not working in IE by using a custom endpoint instead of sending a over-length GET request.

shanewilson:
so this is JUST for IE?

cy:
kind of - it's a workaround for exceeding the GET request length in IE; but all browsers have a GET request limit which might be hit if the facet list gets really long.

cy:
And the list was getting really long because we were sending in the whole facet list - which is silly because the server already knows the whole list. That's why I think it's a good idea even though we have to maintain yet another custom endpoint.

PR #1047, opened 2016-07-22
fix(barChart): PRTL-2 bar stroke complement colour

cy:
screen shot 2016-07-22 at 3 21 30 pm

### DIFF
--- a/app/scripts/components/charts/chart.directives.ts
+++ b/app/scripts/components/charts/chart.directives.ts
@@ -743,20 +743,33 @@ module ngApp.components.charts {
             .attr("y", 0);
 
 
+          let selectionColoursMap = {};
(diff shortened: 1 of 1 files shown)


alex-wilmer:
I think the light border against dark rect like light green / purple is hard to see, but I also don't think it matters at all. If people like it then 👍 

Though I prefer spaces around infix operators `1 + 2` vs `1+2`, not a deal breaker. There's an extra space before `( 1 === 0)`

cy:
yeah, the purple bars are the worst, that's why I did https://github.com/NCI-GDC/portal-ui/pull/1047/files#diff-d5d00df13885fc0ff4b23bde635138b4R754 but it's still not great 😢

seancannata:
@cy This PR can get merged. Thanks!

PR #1050, opened 2016-07-26
feat(header): PRTL-136 login opens on enter

cy:
Tabbing over and pressing enter on Login, Download Token and Logout is the same as clicking

### DIFF
--- a/app/scripts/core/core.directives.ts
+++ b/app/scripts/core/core.directives.ts
@@ -1,5 +1,15 @@
 module ngApp.core.directives {
   import IUserService = ngApp.components.user.services.IUserService;
+  enum KeyCode {
+      Space = 32,
+      Enter = 13,
+      Esc = 27,
+      Left = 37,
+      Right = 39,
+      Up = 38,
+      Down = 40,
+      Tab = 9
+  }
 
   /* @ngInject */
   function loginButton(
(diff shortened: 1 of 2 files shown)


app/scripts/core/core.directives.ts
```diff
@@ -69,6 +80,16 @@ module ngApp.core.directives {
...
+        };
+
+        $element.on('keypress', (e) => {
+          if (e.keyCode === KeyCode.Enter) {
+            openLogin();
+          }
+        });
+
+        $element.on('click', function(){
+          openLogin();
```
alex-wilmer: can this be `$element.on('click', openLogin)` ?

app/scripts/core/core.directives.ts
```diff
@@ -94,8 +115,15 @@ module ngApp.core.directives {
...
-
           $window.location = redirect + authQuery;
+        };
+        $element.on('keypress', (e) => {
+          if (e.keyCode === KeyCode.Enter) {
+            logout();
+          }
+        });
+        $element.on('click',function() {
+          logout();
```
alex-wilmer: same as above

alex-wilmer:
comments aside, 👍

app/scripts/core/core.directives.ts
cy: good point

PR #1053, opened 2016-07-26
feat(customFacets): PRTL-67 filename as id facets

cy:
screen shot 2016-07-26 at 5 28 05 pm

### DIFF
--- a/app/scripts/components/facets/facets.services.ts
+++ b/app/scripts/components/facets/facets.services.ts
@@ -318,7 +318,7 @@ module ngApp.components.facets.services {
       var facetType = 'terms';
       if (_.includes(fieldName, 'datetime')) {
         facetType = 'datetime';
-      } else if (_.some(['_id', '_uuid', 'md5sum'], a => _.includes(fieldName, a))) {
+      } else if (_.some(['_id', '_uuid', 'md5sum', 'file_name'], a => _.includes(fieldName, a))) {
         facetType = 'id';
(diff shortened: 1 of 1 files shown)


alex-wilmer:
👍

app/scripts/components/facets/facets.services.ts
```diff
@@ -318,7 +318,7 @@ module ngApp.components.facets.services {
       var facetType = 'terms';
       if (_.includes(fieldName, 'datetime')) {
         facetType = 'datetime';
-      } else if (_.some(['_id', '_uuid', 'md5sum'], a => _.includes(fieldName, a))) {
+      } else if (_.some(['_id', '_uuid', 'md5sum', 'file_name'], a => _.includes(fieldName, a))) {
```
shanewilson: is this information in the `/mapping` data?

cy: partially, each field does have a `type`. For `type: long`, facet type will always be range, but for `type: string` sometimes (the ones in the array) an id facet is wanted, sometimes it's a date facet, otherwise it's a term facet.

shanewilson: Ok - I think we could actually push all this into the API at some point. But that can wait.

shanewilson:
👍

seancannata:
@cy This can get merged. Thanks! ✅

PR #1054, opened 2016-07-29
feat(app): PRTL-172 legacy data banner

alex-wilmer:
Closes #PRTL-172


alex-wilmer:
@BedfordWest where should I be creating this pull request against?

alex-wilmer:
This should change so that the banner is at initial depth and the rest of the site is translated down by some default. Then when dismissed,a single transform property goes to zero using gpu.

alex-wilmer:
https://github.com/NCI-GDC/portal-ui/pull/1056

PR #1055, opened 2016-08-02
fix(app): PRTL-226 add project info to files

alex-wilmer:
- fix bug where incorrect access in cart

Closes #PRTL-226


shanewilson:
👍

PR #1058, opened 2016-08-04
feat(app): PRTL-172 legacy data banner

alex-wilmer:
Closes #PRTL-172


alex-wilmer:
Would have to es5 some things to get this working. Because we're planning release for Monday this should wait

PR #1063, opened 2016-08-10
fix(app): PRTL-237 banner about portal status

alex-wilmer:
Closes #PRTL-237


BedfordWest:
Verified it looks ok in QA.

PR #1068, opened 2016-08-15
feat(quicksearch): PRTL-29 support annotations

cy:
requires https://github.com/NCI-GDC/gdcapi/pull/577/files
screen shot 2016-08-15 at 5 23 54 pm
only problem is the icon for annotations is a png not a font 😓


shanewilson:
I would just use a font awesome icon for this - until we get the new set of icons

cy:
screen shot 2016-08-17 at 1 51 18 pm

PR #1081, opened 2016-09-09
Feat/278 token

cy:


shanewilson:
does this make a request for the token every time? or just the first time?

cy:
it should make a new request unless it's cached - turning caching off for auth

shanewilson:
rebase like you normally would but use the Squash and Merge button - it will put the PR in the commit header

PR #1082, opened 2016-09-13
feat(annotations): PRTL-87 update

cy:
but some annotations are missing in the annotations index (DAT-168)


shanewilson:
this should probably be QA'ed?

cy:
yeah

PR #1083, opened 2016-09-14
Relay

shanewilson:
TODOs

- [x] Login / Logout
- [x] Home page data portal summary queries
- [x] (Hard coded?) Facets everywhere
   - [x] term
   - [x] suggestion
   - [x] prefix
   - [x] range https://github.com/NCI-GDC/portal-ui/pull/1220
   - [x] date https://github.com/NCI-GDC/portal-ui/pull/1228
   - [x] add date/range to cohort aggs
   - [x] custom facets
   - [x] reset single facet on each facet header
   - [x] search within term list on term facet
   - [x] warning message on age at diagnoses if user enters age >= 90
- [x] Download buttons
- [x] Smart Search page
- [x] Add to cart all / filtered buttons
- [x] Summary section repo page
- [x] Cohort page
- [x] quick search
- [x] loading spinners
- [x] double check 508
- [x] triple check 508
- [x] info / warning modals
- [x] notification banners
- [x] Move all rest calls to relay containers
- [x] annotation entity page (make sure the counts follow these rules https://jira.opensciencedatacloud.org/browse/PRTL-87)
- [x] bam slicing


shanewilson:
? didn't I rebase all this into one commit?

modules/node_modules/@ncigdc/dux/relayLoading.js
```diff
@@ -0,0 +1,20 @@
...
+    case 'TOGGLE_LOADING':
+      return action.payload
+    default:
+      return state
+  }
+}
+
+/*----------------------------------------------------------------------------*/
+
+export default relayLoading
```
shanewilson: is this for table spinners and stuff?

alex-wilmer: It *was* for initial page loading, I just copied over the folder from legacy branch, so will likely remove some of these as relay takes over the need for it

modules/node_modules/@ncigdc/containers/FilesCount.js
```diff
@@ -0,0 +1,40 @@
...
+      ? props.hits.pagination.total.toLocaleString()
+      : <span className="fa fa-spinner fa-spin" />
+    }
+  </span>
+);
+
+export const FilesCountQuery = {
+  fragments: {
+    hits: () => Relay.QL`
+      fragment on FileConnection {
```
shanewilson: Looks like `FilesCount` and `CasesCount` are the same thing. I think it would be easier to combine them by changing the query to 

```
fragment ESPagination {
  total
}
```

and then you can reuse it under any endpoint.

But in general I would leave the GraphQL queries

shanewilson:
Try not to get too granular with the Relay components - doing most of the Querying at the Page level and using dumb Components will make things easier. Then we can create Relay containers for the bigger components or ones we use across many pages. If we start by making everything its own Container I'm afraid we might just get overwhelmed.

shanewilson:
So one rules I've used is if a component uses a specific field like - `case_id` or `primary_site` make it a container with its own Query that asks for those fields - but if it use generic fields like `hits` or `pagination`you can just make it a component and pass data down into it

modules/node_modules/@ncigdc/containers/ProjectTr.js
```diff
@@ -118,7 +118,14 @@ export const ProjectTrComponent = ({ node, index, theme }: TProps) => {
         </CasesLink>
       </Td>
       <Td>
-        <FilesLink>{node.summary.file_count}</FilesLink>
+        <RepositoryFilesLink
+          query={{
+            filters: makeFilter([{ field: 'cases.project.project_id', value: node.project_id }], false),
+            searchTableTab: 'files',
```
shanewilson: `RepositoryFilesLink` doesn't seem great if you need to repeat this

alex-wilmer:
🎉  all the checks! 

good job team 💯

PR #1089, opened 2016-09-28
fix(manifest): API-58 remove fields parameter from manifest download

dankolbman:
Manifest fields are strictly defined within https://github.com/NCI-GDC/gdcapi/blob/fbad1249370c2a7eaa2fb1c5c58292be8fbda75a/gdcapi/download/utils.py

Tested in dev, doesn't seem to break anything, but...

r? @NCI-GDC/oicrdevs


MNadolsky:
✅

PR #1090, opened 2016-10-05
fix(facets): PRTL-290 not cutoff anymore

cy:
Closes #PRTL-290


shanewilson:
this is the same as https://jira.opensciencedatacloud.org/browse/LGCY-25 right?

cy:
yes, same

PR #1092, opened 2016-10-07
feat(file): [PRTL-20] adds index file table

shanewilson:
Closes #PRTL-20

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,13 @@
 .idea/
 .jest/
 .vscode/
+.history/
 
 node_modules/
 bower_components/
 typings/
 npm-debug.log*
+lerna-debug.log
 .DS_Store
 
 .env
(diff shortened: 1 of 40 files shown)


shanewilson:
@cy should I use the `tablesToDisplay` thing? I wasn't really sure what that did.

cy:
is the index file table going to be shown regardless of data_category? then you don't.

shanewilson:
it will always be shown

cy:
looked at the spec, if you scroll over horizontally it says
screen shot 2016-10-17 at 3 31 27 pm but it's based on data_format & data_type so use ng-if here https://github.com/NCI-GDC/portal-ui/pull/1092/files#diff-3acfaf0591334ca8b64227bfc57a39d5R494 and rename `tablesToDisplay` to `tablesToDisplayByCategory`?

PR #1094, opened 2016-10-11
Feat/300 gene pg

cy:


cy:
merged into https://github.com/NCI-GDC/portal-ui/pull/1096

PR #1095, opened 2016-10-13
feat(app): PRTL-295 update error messages

cy:
- remove 'accept certificate' message
- always link gdc support in error modals
  Closes #295

### DIFF
--- a/app/scripts/app.ts
+++ b/app/scripts/app.ts
@@ -184,19 +184,7 @@ function appRun(
       config.apiIsMismatched = true;
     }
   }, function(response) {
-    notify.config({ duration: 60000 });
-    notify.closeAll();
-    notify({
-      message: "",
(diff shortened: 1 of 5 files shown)


app/scripts/core/core.services.ts
```diff
@@ -30,7 +30,7 @@ module ngApp.core.services {
       this.setLoadedState(true);
       Restangular.setErrorInterceptor((response, deferred, responseHandler) => {
         this.xhrDone();
-        if (response.status >= 500) {
+        if (response.status >= 500 || response.status === -1) {
```
alex-wilmer: `-1` !

cy: opps, when testing locally & the api is off response.status is -1. removed

seancannata:
✅ This can get merged.

PR #1100, opened 2016-10-20
Feat/347 cancer distribution

cy:
wip


cy:
screencapture-localhost-3000-genes-ensg00000006283-1477063305513

cy:
screen shot 2016-10-24 at 5 04 29 pm
assumed that all the cases in the case-centric index are the ones that are anaylzed to get ssm. Denominator in '# affected cases in [project]' - agg in case index over project_id, denominator in '# affected cases in all project' - sum of agg in case index over all projects

cy:
screen shot 2016-10-26 at 12 54 32 pm
screen shot 2016-10-26 at 12 54 55 pm
@alex-wilmer used your tooltips. r?

PR #1108, opened 2016-10-27
fix(app): PRTL-374 ie padding issue

cy:
angular in ie wasn't evaluating expression in template style attribute, changed to ng-style - working.


seancannata:
@cy This can get merged when you are ready. Thanks! ✅

PR #1112, opened 2016-10-28
use config.es_host

cy:
but don't have case.project.project_id in rebuilt gene index, gene page & top mutated genes barchart not working


cy:
not sure how to fix the tests after 30min of messing with them 😖

PR #1120, opened 2016-11-02
:bug: project list viz fix ups

cy:
- stacked bar chart:
	* 20 genes instead of 10
	* show 'no data' when loading
	* correct query to pass in project id in es script
	score, filter out cases with other project ids
- piechart: show hand when hovering and has callback
screen shot 2016-11-02 at 12 19 25 pm

### DIFF
--- a/app/scripts/projects/projects.controller.ts
+++ b/app/scripts/projects/projects.controller.ts
@@ -65,15 +65,16 @@ module ngApp.projects.controllers {
       $scope.tableConfig = this.ProjectsTableService.model();
 
       this.refresh();
-      this.renderReact([], []);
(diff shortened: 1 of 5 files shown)


app/scripts/projects/projects.controller.ts
```diff
@@ -65,15 +65,16 @@ module ngApp.projects.controllers {
       $scope.tableConfig = this.ProjectsTableService.model();
 
       this.refresh();
-      this.renderReact([], []);
+      this.renderReact([], [], true);
```
alex-wilmer: Can you make this take an object? signature kinda hard to remember

shanewilson:
so blue!

PR #1125, opened 2016-11-02
:arrow_up: upgrade oncogrid to fix removing track icons

Jephuff:


cy:
w00t can install oncogrid now

PR #1126, opened 2016-11-03
started adding survival legend

Jephuff:


Jephuff:
@alex-wilmer this is the pr with the legend/pValue

PR #1134, opened 2016-11-03
fix survivalplot and oncogrid resizing

Jephuff:


Jephuff:
It might be that you have an outdated survivalplot installed, could you try updating it and see if the problem goes away?

cy:
that worked 👍

PR #1137, opened 2016-11-04
polyfills for IE

Jephuff:
IE wont load because window.location.pathname.includes does not exist.


alex-wilmer:
Doesn't the polyfill add `includes`?

Jephuff:
Sure does, removed that commit 👍

PR #1138, opened 2016-11-04
Prtl 396 support png in ie

Jephuff:


cy:
I'm doing something wrong again? Deleted and re-installed node_modules, getting `aught Error: an HTMLElement or SVGElement is required; got null(…)` trying to dl either SVG on PNG both in chrome & ie

Jephuff:
@cy nope, that was my fault. The selector wasn't working to grab the bar chart. Fixed now.

PR #1140, opened 2016-11-09
[PRTL-390] shorten pValue

Jephuff:


cy:
dumb question - but where's the pvalue shown? I don't see it.
screen shot 2016-11-24 at 2 00 48 pm

Jephuff:
It's only shown when there are 2 curves, so if you click on one the genes it should show up just above the chart.

cy:
I see, I see

PR #1143, opened 2016-11-10
:sparkles: tooltip as react component

cy:
@alex-wilmer wrapping your global-tooltip as a react component like this...?
screen shot 2016-11-10 at 4 53 47 pm

### DIFF
--- a/app/react-components/uikit/Tooltip.js
+++ b/app/react-components/uikit/Tooltip.js
@@ -1,0 +1,26 @@
+// Vendor
+import React, { PropTypes } from 'react';
+
+const Tooltip = ({ innerHTML, children }) => (
+  <span 
+    onMouseOver={() => { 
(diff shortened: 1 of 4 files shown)


alex-wilmer:
haha... this is so unconventional, but yet so clever. I like it

app/react-components/uikit/Tooltip.js
```diff
@@ -0,0 +1,26 @@
+// Vendor
+import React, { PropTypes } from 'react';
+
+const Tooltip = ({ innerHTML, children }) => (
+  <span 
+    onMouseOver={() => { 
+      let tooltip = document.querySelector('.global-tooltip')
+      tooltip.innerHTML = innerHTML;
+      tooltip.className = 'global-tooltip active';
```
Jephuff: Might what to switch this to classList. Probably there will never be another class on the tooltip, but just in case :)

PR #1149, opened 2016-11-17
:sparkles: convert aa change to single letter notation

cy:
none of the the react components had tests yet, pulled in jest; not sure if doing it 100% correctly. Also there was a bunch of protractor tests in `app/tests` that hadn't been touched in 2 years, I deleted them because Jest was trying to run them. Were they being used?


cy:
replaced by https://github.com/NCI-GDC/portal-ui/pull/1151

PR #1151, opened 2016-11-18
:white_check_mark: pull in jest, add makeFilter test

cy:
don't need amino acid notation change any more, pulled jest testing into its own branch and moved the makeFilter test into a jest test. The previous tests had case for 'should double escape if noEscape is false ' https://github.com/NCI-GDC/portal-ui/blob/next/app/scripts/core/tests/makefilter.tests.js#L47 anyone remember why we would need to double escape?


shanewilson:
> anyone remember why we would need to double escape?

nope

PR #1152, opened 2016-11-21
Prtl 445 add dropdown for svg download 

Jephuff:


app/react-components/components/DownloadVisualizationButton.js
```diff
@@ -0,0 +1,89 @@
...
+}
+const styles = {
+  hidden: {
+    width: 0,
+    height: 0,
+    overflow: 'hidden',
+  },
+};
+
+const icon = <i className={"fa fa-download"}/>;
```
cy: for icons, can you use the `react-icons` components as in https://github.com/NCI-GDC/portal-ui/blob/8978e029bb49839c425c9558807608e2c0f7c884/app/react-components/Project.js#L4

app/react-components/uikit/Button.js
```diff
@@ -0,0 +1,47 @@
+// Vendor
+import React from 'react';
+import Radium from 'radium';
```
cy: hang on are we going with Radium or glamor?

app/react-components/Button.js
```diff
@@ -1,68 +0,0 @@
-// Vendor
-import React from 'react';
```
cy:
```Cannot find module '../Button' from '/Users/CYu/repos/portal-ui/app/react-components/oncogrid'
Error: Cannot find module '../Button' from '/Users/CYu/repos/portal-ui/app/react-components/oncogrid'```

File: app/react-components/uikit/Button.js (hunk 2 above)
Comment (Jephuff) [inline app/react-components/uikit/Button.js]: Radium was just what the button was using on legacy. But thats a very good point, I will convert it to glamour.

File: app/react-components/components/DownloadVisualizationButton.js (hunk 1 above)
Comment (alex-wilmer) [inline app/react-components/components/DownloadVisualizationButton.js]: Could we actually try to avoid react-icons altogether? it still doesn't build with `NODE_ENV=production` fluent in expressjs

Comment (Jephuff) [inline app/react-components/components/DownloadVisualizationButton.js]: So no react-icons? is there an alternative, or should I just leave the className?

Comment (alex-wilmer) [inline app/react-components/components/DownloadVisualizationButton.js]: In subui we have an `/icons` folder.. we can add one here then do something like

```
let WhateverIcon = ({ classname, ...props }) => 
  <i className={`${className} fa fa-whatever} {...props} />
```

or yeah just do it inline. the advantage of the named component is in case we decided to switch what `whatever` icon should be, only to need to change it one place.

but yes please no react-icons for now. I have a long running branch in subui to remove them all.

cy: orz i gotta fixup my branches to not use them then

PR #1157, opened 2016-11-22
:sparkle: [PRTL-371] support list

Jephuff:


shanewilson:
also in the project list page

Jephuff:
Yeah, I was just looking at that. It's still angular, so I suppose I should wait until it's react?

shanewilson:
you can make the change in ng since it is so small. the table headers can stay singular.

PR #1159, opened 2016-11-24
Prtl 392 survival to tsv

Jephuff:
based on https://github.com/NCI-GDC/portal-ui/pull/1152


app/react-components/components/SurvivalPlotWrapper.js
```diff
@@ -41,13 +41,29 @@ const styles = {
   },
 };
 
+function toMap(d) {
```
alex-wilmer: Maybe make this a little more general or add an extra argument for sorting?

function toMap (d, sort = sortById) {
  return new Map(Object.entries(d).sort(sort)
}
just a thought

Jephuff:
merged into https://github.com/NCI-GDC/portal-ui/pull/1160

PR #1160, opened 2016-11-25
survivalplot fetching and behaviour updates

Jephuff:
https://github.com/NCI-GDC/gdcapi/pull/634


app/react-components/Project.js
```diff
@@ -395,7 +396,33 @@ const Project = ({
...
+                    if (d.survivalId !== geneData.id) {
+                      const filters = [
+                        {"op": "=","content": {"field": "genes.symbol","value": d.survivalId}},
+                        {"op": "not","content": {"op": "=","content": {"field": "genes.symbol","value": d.survivalId}}}
+                      ];
+
+                      const tempFilters = [
+                        {"op":"=","content":{"field":"cases.project.project_id","value":"TCGA-KIRC"}},
+                        {"op":"=","content":{"field":"cases.project.project_id","value":"TCGA-KIRP"}}
+                      ];
```
shanewilson: are these used?

Jephuff: Yep, temp filters is used in place of the ICGC endpoint until we get the data in the api. I will add a comment to make it a little clearer why it's there though.

app/react-components/Project.js
```diff
@@ -395,7 +396,29 @@ const Project = ({
                 { key: 'num_mutations', title: '# Mutations'},
                 {
                   title: <i className="fa fa-bar-chart-o"><div style={styles.hidden}>add to survival plot</div></i>,
-                  onClick: (d) => setSurvivalGene(d === survivalGene ? null : d),
+                  onClick: (d) => {
+                    if (d.survivalId !== geneData.id) {
+                      const filters = [
+                        // TODO: switch to genes.symbol filters when api is ready.
```
Jephuff: Was just waiting for the data to be there so I could test the queries. Does the elastic 5 update include the data needed?

shanewilson: I forgot to update the collab UI to point to es5 - sec

Jephuff: ```http://localhost:5000/analysis/survival?filters=[{"op":"and","content":[{"op":"=","content":{"field":"cases.project.project_id","value":"TCGA-KIRC"}},{"op":"=","content":{"field":"cases.project.genes.symbol","value":"VHL"}}]}]``` this is returning 0 results. 
Is my query messed up?

shanewilson: are genes under projects?

Jephuff: oops, pasted the wrong one, I tried both though.

shanewilson: doesn't look like genes are in `gdc_from_graph/case`

app/scripts/projects/module.ts
```diff
@@ -179,7 +179,7 @@ module ngApp.projects {
         ): ng.IPromise => {
           return $http({
             method: 'GET',
-            url: `${config.api}/analysis/survival?filters={"op":"=","content":{"field":"cases.project.project_id","value":"${$stateParams["projectId"]}"}}`,
+            url: `${config.api}/analysis/survival?filters=[{"op":"=","content":{"field":"project.project_id","value":"${$stateParams["projectId"]}"}}]`,
```
shanewilson: can you add `cases` back? it makes it so the filters can be reused across different types - doesn't matter here since it is against case but for consistency

Jephuff: How do I make that work? when I have `cases` I get 0 results.

PR #1163, opened 2016-11-30
:sparkles::bug: Update Frequent Mutations

alex-wilmer:


app/react-components/Gene.js
```diff
@@ -179,6 +179,11 @@ let Gene = (() => {
         return {
           ...x,
           num_affected_cases_all: x.occurrence.length,
+          num_affected_cases_by_project: x.occurrence.reduce((acc, o) => ({
+            ...acc,
+            [o.case.project.project_id]: acc[o.case.project.project_id] ? acc[o.case.project.project_id] + 1 : 1
```
shanewilson: I'm not saying do this but does this work?

`[o.case.project.project_id]: (acc[o.case.project.project_id] || 0) + 1`

shanewilson: also might be worth making this a function you can import since you are using it a lot

alex-wilmer: heh that's pretty clever, yeah looks like it would work to me. more readable? I dunno

alex-wilmer:
closing in favor of. I have some fixes that apply to both and the commits are well marked https://github.com/NCI-GDC/portal-ui/pull/1164

PR #1164, opened 2016-11-30
Feat/update lolliplot

alex-wilmer:


shanewilson:
is this ready to merge?

alex-wilmer:
Yes it's ready, with the caveat that there are more tickets relating to frequent mutations since the last demo. It's probably better to have those in separate PRs I guess?

PR #1171, opened 2016-12-05
:bug: null check on cytoband

alex-wilmer:


alex-wilmer:
this is included elsewhere, but in case this can get in first for new PRs

PR #1172, opened 2016-12-05
check user dl permission - match api

cy:
needs cleanup/tests fixing

### DIFF
--- a/app/scripts/components/user/user.services.ts
+++ b/app/scripts/components/user/user.services.ts
@@ -167,8 +167,7 @@ module ngApp.components.user.services {
           return participant.project.project_id;
         }));
       }
-      // [SV-506] return true if file has no projects, defer permission check to ACL instead of projects
-      return projectIds.length === 0 ? true : !!_.intersection(projectIds, this.currentUser.projects.gdc_ids).length;
+      return _.intersection(projectIds, this.currentUser.projects.gdc_ids).length !== 0;
     }
 
     setUserProjectsTerms(terms) {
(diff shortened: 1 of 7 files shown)


app/scripts/components/user/user.services.ts
```diff
@@ -134,16 +132,11 @@ module ngApp.components.user.services {
...
-            if (p.indexOf("_member_") !== -1) {
-              acc.push(key);
-            }
-            return acc;
-          }, [])
-        }
+        projects: Object.keys(user.projects).reduce((acc, k) => {
+          acc[k] = Object.keys(user.projects[k]).filter(project => user.projects[k][project].indexOf("_member_") !== -1)
+          return acc;
+        }, {})
```
shanewilson: so this will have `phs_ids` and `gdc_ids`?

app/scripts/search/search.files.table.model.ts
```diff
@@ -149,7 +149,9 @@ module ngApp.search.models {
       "submitter_id",
       "cases.case_id",
       "cases.project.project_id",
-      "cases.project.name"
+      "cases.project.name",
+      "cases.project.released",
```
cy: True. So I'll remove this check. Was following this logic in the ticket:

```
For download from active/legacy portal, api checks : 
1. `_member_` role phsid_ids has intersection with file.acl 
2. file in the correct state

Definition of file correct state:
- Legacy data: file in the correct state means file.state=='live' 
- active data: file.state == 'submitted'  and file.file_state in ["submitted", "processing", "processed"] and file.project.released is True
```

app/scripts/components/user/user.services.ts
cy: yup

cy:
read the ticket more closely & updated.

cy:
Done messing with this, replaced this strangeness https://github.com/NCI-GDC/portal-ui/pull/1172/commits/c959402fc795237f990c4474618e07d8f4612a88#diff-ad165250a87fb69ad37b790093d0fc73R170 with an `||` here https://github.com/NCI-GDC/portal-ui/pull/1172/commits/473bb79e31fc5af0153e9d402339502c375f5133#diff-ad165250a87fb69ad37b790093d0fc73R205. Don't know what I was thinking before. Added tests for `userCanDownloadFiles`

cy:
to be implement in react ver

PR #1173, opened 2016-12-06
fix(project): dont fail on missing project data

shanewilson:


app/react-components/Project.js
```diff
@@ -211,8 +211,8 @@ const Project = ({
             thToTd={[
               { th: 'Project ID', td: project.project_id },
               { th: 'Project Name', td: project.name },
-              { th: `Disease Type${project.disease_type.length > 1 ? 's' : ''}`, td: project.disease_type.map(p => <div key={p}>{p}</div>) },
-              { th: `Primary Site${project.primary_site.length > 1 ? 's' : ''}`, td: project.primary_site.map(p => <div key={p}>{p}</div>) },
+              { th: 'Disease Type', td: (project.disease_type || []).map(p => <div key={p}>{p}</div>) },
```
alex-wilmer: I think a nicer version of that is `[].concat(project.disease_type).map`

Jephuff: Right now it will break with strings though, because strings don't have `map`

shanewilson: :confounded: oh yea

PR #1183, opened 2016-12-12
:sparkles: [PRTL-49] data_release info from /status

alex-wilmer:
requires https://github.com/NCI-GDC/gdcapi/pull/660

### DIFF
--- a/app/react-components/Home.js
+++ b/app/react-components/Home.js
@@ -1,13 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
-
-const Row = ({ children, style, className }) =>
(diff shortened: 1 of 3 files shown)


app/react-components/Home.js
```diff
@@ -96,256 +96,258 @@ let styles = {
...
+            />
+              <p>GDC cBio Portal</p>
+          </a>
+        </Row>
+      </Column>
     </Column>
-  </Column>
-)
+  )
+};
```
shanewilson: what changed here? looks like there was a formatting change

alex-wilmer: removed implied return

PR #1184, opened 2016-12-12
:bug::pencil2: [PRTL-475] singularize single case

alex-wilmer:


app/react-components/Mutation.js
```diff
@@ -186,6 +186,8 @@ let Mutation = (() => {
         .map(k => ({project_id: k, ...cancerDistData[k]}))
         .sort((a, b) => b.freq - a.freq);
 
+      const distinctCases = sortedCancerDistData.reduce((acc, d) => [...acc, ...d.cases], []);
```
cy: didn't catch your whole verbal conversation, but re-looked at the query and yeah I think this is correct. Pulling case_ids from the occurrence, then making them extra-distinct if they weren't already by using Set(). It's not like the Most Frequently Mutated Genes case counts where some of the cases could have no ssm.

cy:
this one too:
screen shot 2016-12-12 at 2 43 17 pm
this one
screen shot 2016-12-12 at 2 44 56 pm
i have a pr changing that to 'Unique Cases with Mutation Data', so I'll update that to handle singular so we don't overwrite each other

alex-wilmer:
Are you doing both of those? (the ones one the Projects list page, or just one?

cy:
sure I'll do both of those. list page.

cy:
oh wait a sec, but 0 should also have an 's'. But guess it'll not be in the chart if there's 0.

alex-wilmer:
even if it "was in the chart" there would zero surface area to initialize the tooltip

PR #1185, opened 2016-12-13
:sparkles: Pagination uikit components

alex-wilmer:
The purpose of this PR is to provide generic pagination UI elements that can be added to anything that needs pagination.
 
  `withPagination` is function that returns an HOC who may take an options argument to set the default `first` and `offset`.

Components wrapped with the resulting HOC can also update pagination data via `props.setPagination({ offset })`

Components will need to make api requests are a result of these props changes are recommended to do so in `componentWillReceiveProps` and make necessary comparisons there.

Update:

## This is my proposed API:

```
import { PaginationContainer } from 'uikit/pagination'

<PaginationContainer
  total={totalNumOfItems}
  entityType="Cases"
  onChange={pageInfo => /* do whatever */}
>
  <WhateverComponent />
</PaginationContainer>
```

This will wrap `WhateverComponent` with a pagination header `Showing x-offset of total Cases` and the buttons to move pages at the bottom. The container callback will trigger when someone goes to a new page.

### DIFF
--- a/app/react-components/uikit/ButtonGroup.js
+++ b/app/react-components/uikit/ButtonGroup.js
@@ -1,0 +1,32 @@
+// Vendor
+import React, { Children, cloneElement } from 'react';
+import StyleBuilder from 'style-builder';
+
+// Custom
+import { Row } from './Flex';
+
+/*----------------------------------------------------------------------------*/
+
+const ButtonGroup = ({ style, children, ...props }) => (
+  <Row style={style} {...props}>
+    {Children.map(children, (child, i) =>
+      cloneElement(child, {
+        ...child.props,
+        style: StyleBuilder.build({
+          ...child.props.style,
+          borderRadius: // shouldn't have newlines
+            `${!i ? '4px' : '0px'} `
+          + `${i === children.length - 1 ? '4px' : '0px'} `
+          + `${i === children.length - 1 ? '4px' : '0px'} `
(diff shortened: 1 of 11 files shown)


app/react-components/uikit/Pagination.js
```diff
@@ -0,0 +1,132 @@
...
+
+const PaginationControls = props => {
+  const currentPage = (props.offset / props.first) + 1;
+  const totalPages = props.total / props.first;
+  const pageOffset = 10 * Math.floor((currentPage - 1) / 10);
+
+  return (
+    <ButtonGroup>
+      <PaginationBtn onClick={() => props.setPagination({ offset: 0 })}>{'<<'}</PaginationBtn>
+      <PaginationBtn onClick={() => props.setPagination({ offset: props.offset - props.first })}>
```
cy: is there a check for not going negative

app/react-components/uikit/ButtonGroup.js
```diff
@@ -0,0 +1,32 @@
...
+import { Row } from './Flex';
+
+/*----------------------------------------------------------------------------*/
+
+const ButtonGroup = ({ style, children, ...props }) => (
+  <Row style={style} {...props}>
+    {Children.map(children, (child, i) =>
+      cloneElement(child, {
+        ...child.props,
+        style: StyleBuilder.build({
```
cy: interesting

PR #1186, opened 2016-12-13
:memo::art: lint Mutations

alex-wilmer:


app/react-components/Mutation.js
```diff
@@ -1,15 +1,17 @@
+/* global $ */
```
Jephuff: Should probably import jQuery instead of using the global?

app/react-components/Mutation.js
```diff
@@ -57,320 +61,322 @@ let Mutation = (() => {
...
-      );
-    }
-  }
-})()
+        </Column>
+        <Column style={{ ...styles.card, marginTop: '2rem' }}>
+          <ProteinLolliplotComponent
+            gene={gene}
+            $scope={$scope}
+            reset={this.state.ProteinLolliplot.reset}
```
Jephuff: I think this is suppose to be `props.ProteinLolliplot.reset`

alex-wilmer: oh nice good call

app/react-components/Mutation.js
Jephuff: We do have it https://github.com/NCI-GDC/portal-ui/blob/next/package.json#L118
I'm for removing it entirely

alex-wilmer:
@Jephuff r?

app/react-components/Mutation.js
```diff
@@ -57,320 +60,322 @@ let Mutation = (() => {
...
+            height: 450,
+            domainWidth: this.props.$scope.geneTranscript.length_amino_acid,
+            mutationId: this.props.mutation.ssm_id,
+          }),
+        });
+      },
+    }),
+    lifecycle({
+      componentDidMount() {
+        this.props.renderProteinLolliplot();
```
Jephuff: missed this, `renderProteinLolliplot` needs props

Jephuff:
👍 3

PR #1190, opened 2016-12-14
:bug::sparkles: flag to turn off survival in fm

alex-wilmer:


Jephuff:
We should probably match cancer distribution to frequent mutations
image
image

PR #1193, opened 2016-12-15
:recycle: remove duplicate build filters

Jephuff:


shanewilson:
what does this do?

Jephuff:
We just had two functions doing essentially the same thing, so this removed one.

PR #1196, opened 2016-12-15
:sparkles: pagination for frequentlymutatedgenes

alex-wilmer:


app/react-components/components/FrequentlyMutatedGenesContainer.js
```diff
@@ -0,0 +1,88 @@
...
+
+export default
+compose(
+  withState('state', 'setState', { data: { hits: [], total: 0 }, loading: true }),
+  withProps({
+    fetchData: async props => {
+      props.setState(s => ({ ...s, loading: true }));
+
+      const url =
+        `${props.config.es_host}/${props.config.es_index_version}-gene-centric/gene-centric/_search`;
```
Jephuff: we shouldn't be calling `es_host` any more, would we switch this to hit the api?

alex-wilmer: right thanks, will swap

alex-wilmer:
@Jephuff fixed. Though I think maybe `offset` isn't being handled on this endpoint for some reason? `/analysis/top_mutated_genes_by_project` is that possible?

Jephuff:
Yep, you are correct, it is ignoring it. Do we want to change the name to `offset` or leave it as `from` (it's not handling either right now)

Jephuff:
https://github.com/NCI-GDC/gdcapi/tree/feat/analysis-handle-from this branch accepts `from` I can change it to `offset` if we prefer that.

shanewilson:
use whatever the other endpoints use

alex-wilmer:
offset seems to work here on this endpoint `analysis/frequent_mutations_by_project`

shanewilson:
all the main api/ end points use `from` - the analysis endpoints should follow that

Jephuff:
👍 I'll update

PR #1198, opened 2016-12-16
:art: move CancerDistribution into component, match styles

cy:


alex-wilmer:
the component that you named `CancerDistribution` is really a much more generic component, where all of the stuff specific to cancer distribution is still being passed in top level. My suggestion is to rename the more generic component to something like "CardWithTable" or something and then make another component called CancerDistribution that uses this inner component and passes the headers/data.. or just forget about the CancerDistribution component altogether, though it's nice to see thin components at the "page" level imo

cy:
Originally moved into its own component to wrap the pagination around it, but abandoned that for now due to it being only 1 request & not having much data at that. Plus the Cancer Distributions on Gene vs Mutation page differs by one column so those needed to be passed in as well. It's true, I can generalize it more. If we ever decide to put the pagination in, I don't want to have to do this code moving again.

alex-wilmer:
That's fair, and doesn't need to be done in this PR anyways.

PR #1199, opened 2016-12-16
:arrow_up: lolliplot 0.4.1

alex-wilmer:
:bug: [PRTL-508] prevent user-select on drag
:art: lint ProteinLolliplot component

### DIFF
--- a/app/react-components/.eslintrc
+++ b/app/react-components/.eslintrc
@@ -11,6 +11,7 @@
     "fp/no-mutation": 0,
     "fp/no-mutating-methods": 0,
     "fp/no-this": 0,
+    "jsx-a11y/no-static-element-interactions": 0,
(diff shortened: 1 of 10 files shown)


app/react-components/.eslintrc
```diff
@@ -10,6 +10,7 @@
     "fp/no-mutation": 0,
     "fp/no-mutating-methods": 0,
     "fp/no-this": 0,
+    "jsx-a11y/no-static-element-interactions": 0,
```
shanewilson: we should probably keep this? can turn it off in files if needed

alex-wilmer: Yeah okay, i'll put this back and replace it with `<UnstyledButton />` or something

shanewilson:
probably needs an update

PR #1201, opened 2016-12-19
Feat/case react

alex-wilmer:
Some things here were stripped out from previous PRs (eslint, mostly). and I would like to touch up some things included here, but being out of the office for a while I felt it was better to put this work up incase someone wanted to see it / work with it / borrow some ideas.

# Bio Tree

I would very much like to generalize this component. I've spent a tone of time making this component easier to use / reducing complexity, but as of late I'm certain that a top level controller for all recursively drawn nodes is better than attempting to propagate state changes down the tree.

I don't think there's a rush for getting this PR merged, but this implementation of the tree works in case we do indeed need to merge it sooner. 

# Tabs

I think this component is pretty cool, but the styling was done too directly to fit the "facet tabs" looks. I think it should be a much more generic tab component that is easier to style.

# Frequent Mutations

This component was not fully ready when this PR was started so it should make it way into this PR or shortly thereafter

# Moving queries down

I definitely intend to move the queries out of Angular--period. In the Relay branch they are ready, but they are not ready here... it's not too much work to strip out all data fetching / transformation for this particular entity page to make Angular nothing more than glue, which will make it easier to transition to Relay.. or in the latter outcome, be within the React domain that makes a variety of other logic easy to deal with.


shanewilson:
is this done?

alex-wilmer:
already moved to relay

PR #1203, opened 2016-12-22
:sparkles: move es queries to api

Jephuff:


Jephuff:
requires this PR https://github.com/NCI-GDC/gdcapi/pull/667

alex-wilmer:
code looks good.. want to get a deployment up and running before merging if possible. I think also we should note that some things break if the queries don't come back with the expected data. example http://localhost:3000/projects/CCLE-COAD . could be a different pull request for those though

alex-wilmer:
Ok I'm approving this. we're clearly moving in this direction and we're still developing, but this is such a big change and it works locally so..

PR #1213, opened 2017-01-18
fix oncogrid

Jephuff:
Oncogrid doesn't work with d3v4, as a stopgap I made a version of the repo that has d3v3 as a dependency, I'll make a ticket to update oncogrid to work with v4 but I don't think it will be simple.


alex-wilmer:
it works 👍

PR #1217, opened 2017-01-19
change project_id to project_ids

Jephuff:


alex-wilmer:
it's `project_ids` == csv?

Jephuff:
yep

PR #1220, opened 2017-01-19
Feat/573 range facets relay

cy:
~wip~ done


cy:
@NCI-GDC/oicrdevs bugs ironed out, this is ready for review. 'Age at Diagnoses' follows these interesting conversion rules https://jira.opensciencedatacloud.org/browse/PGDC-2398

For example:
Year From: 1 To: 1 -> convert Days From: 365 To: 729
When setting the filter, it will always set with days as units.

PR #1221, opened 2017-01-20
:sparkles: [PRTL-578] Implement GDC Apps Dropdown

SidSyrus:
- GDC Apps Dropdown implementation
- New component `GDCAppsDropdown`


modules/node_modules/@ncigdc/components/GDCAppsDropdown/style.js
```diff
@@ -0,0 +1,57 @@
+// @flow
+
+import { css, arrowIconRule } from 'glamor';
```
alex-wilmer: `arrowIconRule` is not a function provided by glamor, so this will crash 💥 
`insertRule` is what you want here.

modules/node_modules/@ncigdc/uikit/Dropdown.js
```diff
@@ -16,22 +16,28 @@ const Dropdown = ({
   mouseUpHandler,
   children,
   dropdownStyle = {},
+  dropdownClassName = '',
   itemStyle = {},
+  dropdownItemClass = true,
+  buttonWithIcon = null,
```
alex-wilmer: I think `button` is probably better term here. you're passing in an icon from the outside, but this component doesn't care if it's an icon or not

modules/node_modules/@ncigdc/components/GDCAppsDropdown/style.js
SidSyrus: Right! I will make the change.

modules/node_modules/@ncigdc/uikit/Dropdown.js
SidSyrus: Yup make sense. Will rename the prop

alex-wilmer:
cool, just please squash your commits down to a single commit with a relevant message about the work down.

good job on first PR! 👍

SidSyrus:
Ok. Sounds good. Thanks.

PR #1222, opened 2017-01-23
:sparkles: relay progress indicator (kind of buggy)

alex-wilmer:
This needs some work, but it's going in the right direction I think.


alex-wilmer:
Let's do it the "transition immediately way and show mock sections way".

PR #1228, opened 2017-01-24
:sparkles: [PRTL-582] date facet

cy:
@NCI-GDC/oicrdevs ready for review


modules/node_modules/@ncigdc/components/Aggregations/DateFacet.js
```diff
@@ -0,0 +1,142 @@
...
+      backgroundColor: 'white',
+    },
+    header: {
+      color: props.theme.primaryLight1,
+      fontSize: '1.7rem',
+      marginBottom: '0.5rem',
+      cursor: 'pointer',
+      alignItems: 'center',
+    },
+    label: {
```
alex-wilmer: I think this is the case where `styled` comes in handy, so you don't have to weave theme via `withTheme` and you can remove the style out of `render` so they can be reused

```
import styled from '@ncigdc/theme/styled';
const Label = styled.label({
  backgroundColor: ({ theme }) => theme.greyScale5,
  color: ({ theme }) => theme.greyScale2,
  padding: '0.8rem',
  height: '3.4rem',
  border: ({ theme }) => `1px solid ${props.theme.greyScale4}`,
})
```

cy: ok, I'll give it a shot. but the sucky thing about this airbnb `react-dates` datepicker is it doesn't take a style prop, had to create a css file to override their css :/ But for the label, can do.

alex-wilmer: yeah with the third party stuff it's fine. instead of importing a .css file you can do

```
import { insertRule } from 'glamor'
insertRule(`
  .stuff { color: red; }
`)
```

but I'm not sure that's better.. I do it just cause I don't want to leave js land hehe

cy: Using insertRule mine get overridden ;_; 
screen shot 2017-01-25 at 4 41 50 pm

PR #1229, opened 2017-01-25
:sparkles: Dismissible banner with fetching information from endpoint

SidSyrus:
- Added new Modal Component
- Introduced `react-modal` as dev dependency 
- Moved dismissible banner into its own component
- New `bannerNotification` reducer
- Reading banner notifications from `/notifications` endpoint
- Moved Portal root element and children into their own component


modules/node_modules/@ncigdc/dux/bannerNotification.js
```diff
@@ -0,0 +1,62 @@
...
+  id: 0,
+  dismissible: true,
+  message: <DismissibleBanner />,
+}];
+
+const reducer = (state = initialState, action) => {
+  switch (action.type) {
+    case NOTIFICATION_SUCCESS:
+      return [
+        ...action.payload
```
alex-wilmer: can you wrap this in a null check? if bad `data` comes back from the server this crashes

```
...(action.payload || []).filter
```

alex-wilmer:
great thanks! please squash and merge to a single commit. 👍

PR #1230, opened 2017-01-25
:sparkles: update api calls

Jephuff:


Jephuff:
should be merged with https://github.com/NCI-GDC/gdcapi/pull/686

PR #1233, opened 2017-01-26
♻️ get all data and links working on project page and case page

Jephuff:
~wip~


Jephuff:
I moved visualizations to a separate component because I had already gone through and removed it. this way we can just remove that one component when we are ready.

PR #1235, opened 2017-01-27
PRTL-592 ✨ angular smart search in react

cheapsteak:
PRTL-594

added back route `/query/s?query=${encodeURIComponent(string)}&filters=${jsurl.stringify(obj)}`


cheapsteak:
PRTL-594

PR #1245, opened 2017-02-02
:sparkles: prefix facet for case submitter id

cy:
screen shot 2017-02-02 at 5 56 52 pm
filter doesn't seem to work on the cohort page though. works on repository page.


modules/node_modules/@ncigdc/components/Aggregations/PrefixFacet.js
```diff
@@ -0,0 +1,140 @@
...
+  setActive,
+  active,
+  mouseDownHandler,
+  mouseUpHandler,
+}) => {
+  const setQuery = (v) => {
+    console.log(v);
+  };
+  const query = v => ({
+    offset: 0,
```
alex-wilmer: would it be convenient to use this function here? https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/@ncigdc/utils/filters/index.js#L167

cy: yup can, done

alex-wilmer:
awesome

PR #1246, opened 2017-02-03
:bug: use search instead of query

Jephuff:


alex-wilmer:
I just fixed the failing test from before

PR #1247, opened 2017-02-03
✨  [PRTL-628] add sort and download actions to tables

Jephuff:


alex-wilmer:
two main things I would change is to use `Dropdown` instead of `withDropdown`. it’s a little bit more convenient (don’t need to pass mousehandlers / active)

[4:39]  
also that way you can wrap just the button part in a tooltip so it doesn’t stay up on top of the dropdown content

[4:40]  
and just the positioning of the dropdown to be below the button

[4:40]  
other than that. it’s great

alex-wilmer:
nice, looks good 👍

PR #1252, opened 2017-02-07
✨[PRTL-638] entity page actions

Jephuff:
~~wip~~

### DIFF
--- a/modules/node_modules/@ncigdc/components/Case.js
+++ b/modules/node_modules/@ncigdc/components/Case.js
@@ -4,6 +4,7 @@ import React from 'react';
 import FileIcon from 'react-icons/lib/fa/file-o';
 import { compose } from 'recompose';
 import JSURL from 'jsurl';
+import { connect } from 'react-redux';
 
 import LocationSubscriber from '@ncigdc/components/LocationSubscriber';
 import withRouter from '@ncigdc/utils/withRouter';
@@ -20,6 +21,8 @@ import BiospecimenCard from '@ncigdc/components/BiospecimenCard';
 import FrequentMutationsContainer from '@ncigdc/components/FrequentMutationsContainer';
 import { withTheme } from '@ncigdc/theme';
 import Link from '@ncigdc/components/Links/Link';
+import Button from '@ncigdc/uikit/Button';
+import { toggleFilesInCart } from '@ncigdc/dux/cart';
 
 import type { TRawQuery } from '@ncigdc/utils/uri/types';
 
@@ -42,7 +45,8 @@ const styles = {
 
 const enhance = compose(
   withRouter,
-  withTheme
+  withTheme,
+  connect(state => ({ cartFiles: state.cart.files }))
 );
 
 const Case = ({
@@ -52,6 +56,10 @@ const Case = ({
   numCasesAggByProject,
   push,
   query,
+  dispatch,
+  cartFiles,
+  totalFiles,
+  files,
 }: {
   node: Object,
   totalNumCases: number,
@@ -59,6 +67,10 @@ const Case = ({
   numCasesAggByProject: Array<{ [string]: number }>,
   push: Function,
   query: Object,
+  dispatch: Function,
+  cartFiles: Array<Object>,
+  totalFiles: number,
+  files: Array<Object>,
 }) => {
   const p = node;
 
@@ -156,8 +168,16 @@ const Case = ({
     });
   }, []);
 
+  const filesToAdd = files.filter(f => !cartFiles.some(cf => cf.file_id === f.file_id));
+
   return (
     <Column spacing={theme.spacing}>
+      <Row style={{ justifyContent: 'flex-end' }}>
+        <Button onClick={() => dispatch(toggleFilesInCart(filesToAdd.length ? filesToAdd : files))}>
+          {filesToAdd.length ? 'Add all files to the cart' : 'Remove all files from the cart'}
+        </Button>
+      </Row>
+
       <Row spacing={theme.spacing}>
         <EntityPageVerticalTable
           id="summary"
@@ -179,10 +199,11 @@ const Case = ({
 
         <Column style={{ width: '200px' }} spacing={theme.spacing}>
           <CountCard
+            style={{ width: 'auto' }}
             title="FILES"
-            count={p.files.hits.total.toLocaleString()}
+            count={totalFiles.toLocaleString()}
             icon={<FileIcon style={styles.icon} />}
-            linkParams={p.files.hits.total ? {
+            linkParams={totalFiles ? {
               pathname: '/repository',
               query: {
                 filters: makeFilter([{ field: 'cases.case_id', value: p.case_id }], false),
@@ -192,6 +213,7 @@ const Case = ({
             } : null}
           />
           <CountCard
+            style={{ width: 'auto' }}
             title="ANNOTATIONS"
             count={(annotationIds || []).length.toLocaleString()}
             icon={<i className="fa fa-edit" style={styles.icon} />}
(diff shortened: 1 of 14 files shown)


modules/node_modules/@ncigdc/components/Project.js
```diff
@@ -119,22 +105,33 @@ const Project = ({
         <DownloadButton
           disabled={!biospecimenCount}
```
cy: screen shot 2017-02-10 at 10 47 02 am
grey these buttons out when they're disabled?

modules/node_modules/@ncigdc/components/Case.js
```diff
@@ -156,8 +164,16 @@ const Case = ({
     });
   }, []);
 
+  const filesToAdd = p.files.filter(f => !files.some(cf => cf.file_id === f.file_id));
+
   return (
     <Column spacing={theme.spacing}>
+      <Row style={{ justifyContent: 'flex-end' }}>
+        <Button onClick={() => dispatch(toggleInCart(filesToAdd.length ? filesToAdd : p.files))}>
```
cy: I get 
```Uncaught TypeError: (0 , _cart.toggleInCart) is not a function
    at onClick (http://localhost:8080/js/bundle.js:148054:52)
```
on this button, plus I don't see it in `dux/cart.js` maybe it got renamed to `toggleFilesInCart` behind your back

modules/node_modules/@ncigdc/components/Project.js
Jephuff: 👍 for see any issues with me making that global on the Button component? so if you pass in `disabled` it will disable hover and change it to grey?

alex-wilmer: I thought we had that already! Yes please do that

cy:
while you're here could you add a null check to https://github.com/NCI-GDC/portal-ui/pull/1252/files#diff-a910d5d773a62373e28cd4a4fd61c612R111 and https://github.com/NCI-GDC/portal-ui/pull/1252/files#diff-a910d5d773a62373e28cd4a4fd61c612R135 otherwise can't switch to that tab if there are no treatments with `Uncaught TypeError: Cannot read property 'length' of undefined`

modules/node_modules/@ncigdc/components/Case.js
Jephuff: you are correct, fixed.

modules/node_modules/@ncigdc/components/File.js
```diff
@@ -54,19 +65,36 @@ const File = ({ node, theme, searchTerm, setState }) => {
...
+          leftIcon={<ShoppingCartIcon />}
+        >
+          {fileInCart(files, node) ? 'Remove from Cart' : 'Add to Cart'}
+        </Button>
+        {
+          userCanDownloadFile({ user, file: node }) ? (
+            <DownloadButton
+              extraParams={{ ids: node.file_id }}
+              filename={node.file_name}
+              url={`${API}data?annotations=true&related_files=true`}
```
cy: I think for downloading files, should use the auth api endpoint, otherwise controlled files wouldn't get work. Like this: https://github.com/NCI-GDC/portal-ui/pull/1254/files#diff-3a31735e0cbeacf6375e5379ae9742aaR114, I added the GDC_AUTH env var on my PR

Jephuff: so this? `url={`${GDC_AUTH}api/data?annotations=true&related_files=true`}`
or do I need to add `/v0`?

modules/node_modules/@ncigdc/components/Header.js
```diff
@@ -28,7 +28,7 @@ const styles = {
 const Header = compose(
   connect(state => ({
     notifications: state.bannerNotification,
-    loggedIn: state.auth.isAuthorized,
+    loggedIn: state.auth.user && Object.keys(state.auth.user.projects.gdc_ids || {}).length > 0,
```
alex-wilmer: oh you don't need that projects check here. on the portal you're either logged in or you're not. submission-ui had more specifics than that

Jephuff: ahh ok, so this would just be user then. 👍

Jephuff:
This PR is read for review again.

modules/node_modules/@ncigdc/components/File.js
cy: oh sorry didn't see this comment. & correct

modules/node_modules/@ncigdc/components/ClinicalCard.js
```diff
@@ -8,39 +8,84 @@ import Tabs from '@ncigdc/uikit/Tabs';
...
+                op: 'in',
+                content: {
+                  field: 'cases.case_id',
+                  value: [caseId],
+                },
+              },
+            ],
+          }}
+          fields={['case_id']}
+          dataExportExpands={['emographic', 'diagnoses', 'diagnoses.treatments', 'family_histories', 'exposures']}
```
cy: `emographic`?

Jephuff: oops, fixed.

modules/node_modules/@ncigdc/components/ClinicalCard.js
```diff
@@ -8,39 +8,84 @@ import Tabs from '@ncigdc/uikit/Tabs';
...
           ]}
           style={{ flex: 1 }}
         />
       }
       {activeTab === 1 &&
         <SideTabs
           contentStyle={{ border: 'none' }}
-          tabs={(p.diagnoses || []).map(x => <p key={x.diagnosis_id}>{x.diagnosis_id}</p>)}
-          tabContent={(p.diagnoses || []).map(x =>
+          tabs={diagnoses.map(x => <p key={x.diagnosis_id}>{x.diagnosis_id}</p>)}
```
cy: seeing `diagnoses.map` is not a function, looks like the array got nested in an inner hits 
screen shot 2017-02-15 at 2 32 48 pm

alex-wilmer: updating to the latest API might fix that? @shanewilson

shanewilson: it should be under `hits` - all nested fields (except like 2) have `hits` now

Jephuff: updated. Haven't moved it to be a container yet, but it handles the data structure.

modules/node_modules/@ncigdc/components/Case.js
```diff
@@ -156,8 +168,16 @@ const Case = ({
...
   }, []);
 
+  const filesToAdd = files.filter(f => !cartFiles.some(cf => cf.file_id === f.file_id));
+
   return (
     <Column spacing={theme.spacing}>
+      <Row style={{ justifyContent: 'flex-end' }}>
+        <Button onClick={() => dispatch(toggleFilesInCart(filesToAdd.length ? filesToAdd : files))}>
+          {filesToAdd.length ? 'Add all files to the cart' : 'Remove all files from the cart'}
+        </Button>
```
cheapsteak: Is it possible for this to be a problem if "Add all files to cart" contains files that were already in the cart (e.g. individually added, added from a different cases, etc)?

modules/node_modules/@ncigdc/components/Modals/NoAccessModal.js
```diff
@@ -1,21 +1,38 @@
...
 
-type TProps = {
-  closeModal: Function,
-};
-
-const NoAccessModal = ({
-  closeModal,
-}: TProps) => (
-  <div>
-    jeffrey's making this, overwrite on merge
```
cy: 😺

modules/node_modules/@ncigdc/components/Case.js
Jephuff: filesToAdd is filtered so it is only the ones that aren't in the cart.
So if some are in the cart, it will add the rest. If none are in the cart, it will remove them all :)

PR #1253, opened 2017-02-10
UAT fixes

Jephuff:


Jephuff:
Set was broken :) it would return [Set] instead of spreading.

alex-wilmer:
Oh right, I remember running into that once.

PR #1255, opened 2017-02-10
💄 widen quicksearch box wider so ‘h’ in "Quick Search" isn’t cut off

cheapsteak:
image


alex-wilmer:
👍

shanewilson:
can you make it bigger? wider and taller? just so the text doesn't look so squished?

cheapsteak:
@shanewilson how's this?
image

alex-wilmer:
@shanewilson can we merge this? do you have a totally different vision?

shanewilson:
I have a totally different vision for quick search in general but this is good until we address that ticket

alex-wilmer:
thanks, cc @cheapsteak

PR #1257, opened 2017-02-13
✨ [PRTL-652] add commit and tag to footer

Jephuff:


Jephuff:
this + https://github.com/NCI-GDC/tungsten/pull/526 will add the commit hash

modules/node_modules/@ncigdc/dux/versionInfo.js
```diff
@@ -1,48 +1,33 @@
 // @flow
-/* global __GDC_API__, __COMMIT_HASH__, __VERSION_TAG__, __COMMIT_LINK__ */
-
 import { fetchApi } from '@ncigdc/utils/ajax';
 import { handleActions } from 'redux-actions';
-import packageJSON from './package.json';
+import packageJSON from '../../../../package.json';
```
shanewilson: this won't work once deployed

shanewilson: can stick with just the commit - I think the best solution is to check if a commit is a tag and then use the tag and if not show the commit. I just noticed that having a tag and a commit hash is confusing to people. but that can happen later

Jephuff: no? why not? the path? or bundling the package.json?

shanewilson: package.json doesn't get bundled and browsers can't access the file system

alex-wilmer: Oh I thought maybe we had a json loader in our webpack pipeline. Would be convenient to have no?

alex-wilmer: it works like this for submission ui

Jephuff: Why wouldn't it be bundled if it's imported? 
Isn't deploying just using the build script in the package.json?

Jephuff:
👍  updated

PR #1259, opened 2017-02-13
PRTL-624 ✨ angular githut in react

cheapsteak:
image

git hut

### DIFF
--- a/modules/node_modules/@ncigdc/components/GitHut.js
+++ b/modules/node_modules/@ncigdc/components/GitHut.js
@@ -70,19 +70,29 @@ class GitHut extends React.Component {
   }
 
   handleAngularRouterRequestGo = (state, params, options) => {
-    const stateTabMap = ({
-      'search.participants': 'cases',
-      'search.files': 'files',
-    });
-
-    this.props.push({
-      pathname: '/repository',
-      children: 'repository',
-      query: {
-        searchTableTab: stateTabMap[state],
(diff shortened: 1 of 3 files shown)


Jephuff:
Click events aren't working for me, is it possible to pass handlers in?

cheapsteak:
Ah good catch, I didn't realize it was clickable
It seems to directly update location, but I neutred angular's location.,... I think I can fix that  :)

modules/node_modules/@ncigdc/components/GitHut.js
```diff
@@ -64,6 +69,22 @@ class GitHut extends React.Component {
     }
   }
 
+  handleAngularRouterRequestGo = (state, params, options) => {
```
cheapsteak: I am open to suggestions for what to name this
It handles when angular's router requests a ui-router state change

cheapsteak:
@Jephuff the hut(s) are now clickable!
We're now hijacking ui-router's $state.go and when the angular app wants to go somewhere, we're handling the request and dispatching transformed params to react-router :D

Jephuff:
hmm, I'm getting an error because params.filter is not defined in `handleAngularRouterRequestGo`?

cheapsteak:
Having trouble reproducing, could you please dump the stacktrace?

cheapsteak:
Should be fixed! Now accounting for clicking on project ids

Jephuff:
nice 👍

PR #1261, opened 2017-02-14
🐛 remove custom payload

Jephuff:
I think this should help with the user errors :)


alex-wilmer:
should make sure this still works. can't remember why the api would be returning text here. maybe we're just missing `accept` here https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/@ncigdc/utils/ajax/index.js#L11

Jephuff:
definitely we need to test. If for what ever reason we need to handle text, we can always just add some error handling in that spot :)
I don't have credentials to log into dev 2 current, but when I do I will test.

Jephuff:
alternative https://github.com/NCI-GDC/portal-ui/pull/1262

PR #1263, opened 2017-02-14
Oncogrid updates

Jephuff:


modules/node_modules/@ncigdc/components/modals/SelectModal.js
```diff
@@ -0,0 +1,95 @@
...
+        options.map((track, i) => (
+          <Row key={track.name}>
+            <input
+              value={i}
+              type="checkbox"
+              name={track.name}
+              id={track.name}
+              checked={selectedOptions.indexOf(i) >= 0}
+              onChange={() => {
+                const index = selectedOptions.indexOf(i);
```
cheapsteak: I can't seem to uncheck a track (clicking once selects a track, clicking twice does not de-select it)

I wonder if it's due to not checking for the value here and triggering on "onChange"

modules/node_modules/@ncigdc/components/modals/SelectModal.js
```diff
@@ -0,0 +1,95 @@
...
+
+                if (index >= 0) {
+                  setSelectedOptions([
+                    ...selectedOptions.slice(0, index),
+                    ...selectedOptions.slice(index),
+                  ]);
```
cheapsteak: It looks like this is not returning the item without the selectedOption

How would you feel about replacing the entirity of the onChange with 

```js
onChange={() => setSelectedOptions(_.xor(selectedOptions, [i])) }
```
?

Jephuff: oops, missed a plus 1, I'll update :)

cheapsteak:
LGTM! 🙌

PR #1268, opened 2017-02-16
✨ [PRTL-420] update download permission check for files with no projects

cy:
ticket says all `Data Type = Copy Number QA Metrics` files will have no projects associated, check file state and file acl instead. There doesn't seem to be any files of that type indexed currently, so wrote some tests.

### DIFF
--- a/modules/node_modules/@ncigdc/containers/FilePage.js
+++ b/modules/node_modules/@ncigdc/containers/FilePage.js
@@ -48,6 +48,8 @@ export const FilePageQuery = {
         experimental_strategy
         md5sum
         state
+        file_state
+        acl
         analysis {
           analysis_id
           workflow_type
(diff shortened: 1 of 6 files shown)


cy:
btw re-install node_modules to get the new tests to run because babel-jest was added

Jephuff:
The logic here looks good to me. Is there a way I can test this? I don't have any projects :(

cy:
@Jephuff no projects? 😢 Ask @Fgerthoffert to email dbGap?

Jephuff:
for the slash on GDC_AUTH could you use `url-join` instead? It will make it a bit more resilient :)

cy:
@Jephuff reseting the GDC_AUTH slash change because you've fixed it in a different pr

shanewilson:
this also seems important - is it ready to merge?

cy:
@shanewilson yup, just tested it again. Should be good.

PR #1272, opened 2017-02-17
🐛 [PRTL-627] Biospecimen detail section

SidSyrus:
- GraphQL query update to get fields mentioned in spec
- Passing type of entity when it is selected


cheapsteak:
@SidSyrus This branch cannot be rebased due to conflicts

PR #1273, opened 2017-02-17
 ✨ [PRTL-595] Cart functionality implementation

SidSyrus:
- Metadata, Manifest, Cart download actions
- Update to download utility with notification
- ESLint error fixes,
- File table UI update

Creating a new PR as branch name changed


modules/node_modules/@ncigdc/components/CartDownloadDropdown.js
```diff
@@ -0,0 +1,161 @@
...
+    <Dropdown
+      dropdownStyle={{
+        marginTop: '2px',
+        borderRadius: '4px',
+      }}
+      dropdownItemClass={false}
+      button={
+        <Button
+          style={{ marginLeft: '10px' }}
+          leftIcon={<DownloadIcon />}
```
cy: Not sure if you saw my reply to your comment in the previous PR but
6450868e-f521-11e6-9ac7-2c250b17ee27
so the spinner's in both places

SidSyrus: I wouldn't put in both places because I won't be able to achieve same behavior with Manifest download.

cy: that's not good, the Manifest download has the same behaviour on the angular ver & ppl complained a lot about missing download spinners previously. 🤔 Also when the item is clicked, doesn't the dropdown close so the spinner can't be seen anyway? Maybe pass down the dropdown's setActive to Manifest's DownloadButton as an optional param `SetParentActive` and set that true/false with DownloadButton's internal setActive? Might work?

SidSyrus: Sure I cam give it a shot. It sounds like a hack though.

modules/node_modules/@ncigdc/components/CartDownloadDropdown.js
```diff
@@ -0,0 +1,161 @@
...
+          >
+            <span>Download {files.length - controlledFiles.length} authorized files</span>
+          </Button>
+        </div>
+      </Column>
+    ));
+  } else {
+    dispatch(setModal(null));
+
+    setActive(true);
```
cheapsteak: Clicking on Download -> Cart
image
Results in 
```
CartDownloadDropdown.js:93 Uncaught TypeError: setActive is not a function
    at downloadCart (http://local.dcc.icgc.org:8080/js/bundle.js:242131:5)
```


I think this function should maybe be in a handler so you don't have to keep track of all the incidental props you need to pass in the places where it's used

https://github.com/acdlite/recompose/blob/master/docs/API.md#withhandlers

SidSyrus: I am not seeing the error on my side.

modules/node_modules/@ncigdc/components/GitHut.js
```diff
@@ -29,8 +29,13 @@ class GitHut extends React.Component {
     const self = this;
     angular
       .module('legacyAngularWrapper', ['ngApp'])
-      .config(($locationProvider) => {
+      .config(($locationProvider, RestangularProvider) => {
         $locationProvider.html5Mode(false);
+        RestangularProvider.setBaseUrl(API);
```
cy: is this a merge error? This PR isn't changing anything in GitHut is it?

SidSyrus: Nop. Its coming form #1276

cy:
wasn't sure how to explain it so pushed a commit that'll show the spinner when re-opening the dropdown. except now I'm not sure how to override the `disabled` state coming from styled orz

modules/node_modules/@ncigdc/components/CartDownloadDropdown.js
```diff
@@ -24,15 +24,19 @@ import download from '@ncigdc/utils/download';
...
     ':hover': {
       backgroundColor: theme.greyScale6,
-    },
+    }
+  }),
+  button: theme => ({
+    borderRadius: '0px',
+    marginLeft: '0px',
+    ...styles.common(theme),
+    '[disabled]': styles.common(theme),
```
cy: 👏 👏

modules/node_modules/@ncigdc/components/Modals/NoAccessModal.js
```diff
@@ -4,35 +4,46 @@ import { connect } from 'react-redux';
...
+  primaryButton,
 }: {
   dispatch: Function,
   message: string,
 }) => (
-  <span>
-    <h2 style={{ paddingLeft: 15 }} >Access Error</h2>
-    <div
+  <Column
+    style={{ margin: '-15px' }}
```
Jephuff: I think if we want those lines to go all the way across it would be better to remove the padding from the modalContainer. Each modal can add their padding then.

modules/node_modules/@ncigdc/containers/FileTable.js
```diff
@@ -28,7 +28,13 @@ export const FileTableComponent = (props: TTableProps) => {
 
   return (
     <div>
+
       <Row style={{ backgroundColor: 'white', padding: '1rem', justifyContent: 'space-between' }}>
+        {props.isCartPage &&
```
Jephuff: We could make this a little more generic, like `tableTitle` and then the cart page can pass in `Cart Items`

modules/node_modules/@ncigdc/containers/FileTable.js
```diff
@@ -94,13 +100,18 @@ export const FileTableComponent = (props: TTableProps) => {
                   padding: '3px',
                 }}
               >
-                <AddToCartButtonAll
-                  edges={props.hits.edges.map(e => e.node)}
-                  total={props.hits.total}
-                />
+                {!props.isCartPage &&
```
Jephuff: And this could change to something like `canAddToCart`

modules/node_modules/@ncigdc/containers/FileTable.js
```diff
@@ -94,13 +100,18 @@ export const FileTableComponent = (props: TTableProps) => {
                   padding: '3px',
```
Jephuff: I don't think this is necessary, should have `padding: 3px` already

modules/node_modules/@ncigdc/components/DownloadButton.js
```diff
@@ -58,6 +67,8 @@ const DownloadButton = ({
           params,
           url,
           method: 'POST',
+          altMessage,
+          dispatch,
```
Jephuff: Doesn't look like the download function uses dispatch? could remove the connect

modules/node_modules/@ncigdc/components/Modals/NoAccessModal.js
Jephuff: https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/uikit/Modal.js#L24
We can remove this padding, and then each modal can add what they need.

SidSyrus: Doesn't work. I tried. It adds `padding:20px` from `react-modal`
https://github.com/reactjs/react-modal#styles

modules/node_modules/@ncigdc/containers/FileTable.js
SidSyrus: What would be advantage of introducing more props when one prop `isCartPage` determines multiple factors?

modules/node_modules/@ncigdc/containers/FileTable.js
SidSyrus: referring back to my other comment.
https://github.com/NCI-GDC/portal-ui/pull/1273#discussion_r103765514

modules/node_modules/@ncigdc/components/Modals/NoAccessModal.js
Jephuff: just pass 0 in for the padding :)

modules/node_modules/@ncigdc/containers/FileTable.js
Jephuff: Because the FileTable probably shouldn't know about the Cart. You want to add features, it's very possible that someone will want to use the Filetable with a title and the ability to add files to cart.

cheapsteak: Alternative solutions: when that happens, e.g. in a project entity page, we can just rename this variable to be `isCartOrProjectPage`. If another page needs it, we can rename it to `isCartOrProjectOr[OtherPage]Page`

Alternatively we can designate `isCartPage` to really mean "should show add to cart button" and either add that as a comment or stick it in our README

modules/node_modules/@ncigdc/components/DownloadButton.js
```diff
@@ -28,14 +31,20 @@ const DownloadButton = ({
   filters = {},
   active,
   setActive,
+  altMessage = false,
+  style = {},
   extraParams = {},
+  dispatch,
 }) => {
-  const icon = <i key="icon" className={`fa fa-download ${active ? 'fa-spinner fa-pulse' : ''}`} />;
+  const icon = active ? <Spinner /> : <DownloadIcon />;
```
Jephuff: These need keys

modules/node_modules/@ncigdc/containers/FileTable.js
Jephuff: I would say the problem is more that `isCartPage` really means `canAddToCart` and `Have a Title` 
those are pretty separate ideas no?

modules/node_modules/@ncigdc/components/Modals/NoAccessModal.js
SidSyrus: aah yes! I could do that.

modules/node_modules/@ncigdc/containers/FileTable.js
SidSyrus: I see. 
so `isCartPage` should be `canAddToCart`
and a new prop for the header.

modules/node_modules/@ncigdc/components/DownloadButton.js
SidSyrus: Whats the purpose of it?

modules/node_modules/@ncigdc/containers/FileTable.js
Jephuff: Yeah, that makes sense to me 👍

modules/node_modules/@ncigdc/components/DownloadButton.js
Jephuff: when ever components are in an array in react they need a unique `key` prop so it knows when they change. (it's put in an array here https://github.com/NCI-GDC/portal-ui/blob/6a5d2b1d1e40b4ecc4d11b816a2e7c7d4fe04a2c/modules/node_modules/%40ncigdc/components/DownloadButton.js#L77 )

cheapsteak: I would personally prefer this

```js
const Icon = active ? Spinner : DownloadIcon;
```

Then instead of 
```js
{text || [icon, <Hidden key="hidden">download</Hidden>]}
```
Have
```js
{text || [<Icon key="icon"/>, <Hidden key="hidden">download</Hidden>]}
```

And instead of `leftIcon={text && icon}` have `leftIcon={text && <Icon />}`

SidSyrus: is `Hidden` for the screen readers? If so why can't we use `aria-label`?

cheapsteak: `aria-label` should also work, although that would go on the button and not the icon
As a tech director of ours was fond of saying, there's more than one way to skin a cat :)

SidSyrus: Not necessarily. To quote Mozilla

> This attribute can be used with any typical HTML element; it is not limited to elements that have an ARIA role assigned.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

Jephuff: Yeah, it's for screen readers. If you test with the wave tool and adding the aria-label works as well, I have no problem with that :) important part is there are no errors in the wave tool.

cheapsteak: That's not why am suggesting to put it on the button

If it helps, think of the icon as a "visual label" for sighted people, telling them "that is what this button does"

modules/node_modules/@ncigdc/components/DownloadButton.js
```diff
@@ -28,14 +31,20 @@ const DownloadButton = ({
   filters = {},
   active,
   setActive,
+  altMessage = false,
+  style = {},
   extraParams = {},
+  dispatch,
```
Jephuff: You can remove dispatch and `connect` completely from this file as they aren't used any more

modules/node_modules/@ncigdc/uikit/Modal.js
```diff
@@ -21,7 +21,7 @@ const modalStyles = {
     background: 'rgb(255, 255, 255)',
     borderRadius: '4px',
     margin: '30px auto',
-    padding: '15px',
+    padding: '0px',
```
Jephuff: Make sure you update everywhere that Modal and setModal is used.

modules/node_modules/@ncigdc/components/DownloadButton.js
cheapsteak: This works:
```html
<button aria-label="download">
   <div class="fa fa-download"></div>
</button>
```

This also works
```html
<button aria-label="download">
   <div class="fa fa-download"></div>
   <div class="visually-hidden">Download</div>
</button>
```

This does _not_ work (wave shows a red "empty button" error)

```html
<button>
   <div aria-label="download" class="fa fa-download"></div>
</button>
```

The aria-label either needs to be on the button, or the button needs to have text to explain what it does

Quoting the MDN link, "The aria-label attribute is used to define a string that labels the **current element**" (emphasis mine). The "current element" is the one that receives focus when tabbed through the page.

The aria label for the icon would also technically be "download icon", but since people using screenreaders wouldn't care about the existence of an icon whose sole existence is to inform sighted users of the purpose of the button, it'd be preferable to leave that unlabled (if it were an img tag instead of just a blank div, it'd be preferable to set it as aria-hidden)

SidSyrus: If you are suggesting icons are only for sighted people to show the purpose of the button then I think I disagree. I have seen multiple sites that use icons as action buttons and in those cases doing something like this would not be wrong.

`<i class="fa fa-undo" aria-label="Undo" onClick="undo()"></i>`

In this particular case however I am suggesting adding 'aria-label' on the button and not on the icon of course. With that we don't have to create a new component `Hidden` which I think is unnecessary when we have HTML core attribute to help us with accessibility.

modules/node_modules/@ncigdc/uikit/Modal.js
Jephuff: There are atleast two more in `containers/CaseAggregations.js` and `containers/FileAggregation.js`

modules/node_modules/@ncigdc/components/DownloadButton.js
cheapsteak: I was of course referring to this icon in this button in this pull request, not making a blanket statement on all the icons on the internet :) 

"Buttons" and "Icons" are separate ideas, just as "Button" and "Text" are separate ideas. Some icons happen to also be buttons, some icons are just icons that happen to be inside buttons. Same with text (spans).

- Visible text labels inside a button labels what the button does to both sighted people and people using screen readers

- Invisible text inside a button (which is also a [perfectly](http://webaim.org/techniques/css/invisiblecontent/) [valid](http://a11yproject.com/posts/how-to-hide-content/) [method](https://snook.ca/archives/html_and_css/hiding-content-for-accessibility) for providing accessibility)  and aria-labels both communicate what a button does to only sighted people

- Icons communicate what a button does to only visual people

A button can have any one or all of those attributes. In our case, regardless of whether you go with

```html
<button>
   <div class="fa fa-download"></div>
   <div class="visually-hidden">Download</div>
</button>
```
or
```
<button aria-label="download">
   <div class="fa fa-download"></div>
</button>
```

This particular icon serves only 1 purpose in both instances (a visual label for sighted users) and would be meaningless for people using screen readers, that is what I meant :)

Your objection to my comment about how the `aria-label` would go on the button and not the icon led me to think you had intended to put it on the icon. If you see benefit in using `aria-label`s instead of visibly hidden text, and if we both agree it should go on the button, then let's put it on the button 😄

PR #1276, opened 2017-02-21
🐛 Set Restangular's base url to "API". Intercept calls to "from=1", change to "from=0"

cheapsteak:


Jephuff:
hmm, I'm only seeing 2 projects on the githut chart but there are 3 on the table. Are they being filtered in some way?

PR #1277, opened 2017-02-21
✨ [PRTL-640] Add Case Files to Cart Button

cheapsteak:
@NCI-GDC/oicrdevs ready for review

### DIFF
--- a/modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
+++ b/modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
@@ -3,7 +3,7 @@
 import React from 'react';
 import _ from 'lodash';
 import { connect } from 'react-redux';
-import { compose, withState, withHandlers } from 'recompose';
+import { compose, withState, withProps, withHandlers } from 'recompose';
 
 import Dropdown from '@ncigdc/uikit/Dropdown';
 import ShoppingCartIcon from 'react-icons/lib/fa/shopping-cart';
@@ -23,8 +23,14 @@ const AddCaseFilesToCartButton = compose(
   connect(state => ({ cart: state.cart })),
   withState('isLoading', 'setIsLoading', false),
   withFilters(),
+  withProps(({ filters }) => ({
+    isFilteringOnFiles: filters && _.some(filters, (item) => {
+      const content = item.hasOwnProperty('content') ? item.content : item;
+      return content.field.indexOf('files.') === 0;
+    }),
+  })),
   withHandlers({
-    handleDropdownActivate: ({ relay, setIsLoading, filters, hasFiles }) => () => {
+    handleDropdownActivate: ({ relay, setIsLoading, filters, isFilteringOnFiles, hasFiles }) => () => {
       if (!hasFiles) {
         // this case has no files to fetch.
         return;
@@ -32,11 +38,11 @@ const AddCaseFilesToCartButton = compose(
       setIsLoading(true);
       relay.setVariables({
         isFileDataRequired: true,
-        filesFilters: filters,
+        filesFilters: isFilteringOnFiles ? filters : null,
       }, (readyState) => _.some([readyState.ready, readyState.aborted, readyState.error]) && setIsLoading(false));
     },
   })
-)(({ hasFiles, files = [], filteredFiles = [], dispatch, handleDropdownActivate, isLoading }) => (
+)(({ hasFiles, files = [], filteredFiles = [], dispatch, handleDropdownActivate, isLoading, isFilteringOnFiles }) => (
   <Dropdown
     dropdownStyle={styles.dropdownContainer}
     dropdownClassName={isLoading ? 'hidden' : 'dropdown-menu'}
--- a/modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
+++ b/modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
@@ -1,0 +1,81 @@
+/* @flow */
+
+import React from 'react';
+import _ from 'lodash';
+import { connect } from 'react-redux';
+import { compose, withState, withHandlers } from 'recompose';
+
+import Dropdown from '@ncigdc/uikit/Dropdown';
+import ShoppingCartIcon from 'react-icons/lib/fa/shopping-cart';
+import TrashIcon from 'react-icons/lib/fa/trash';
+import { Tooltip } from '@ncigdc/uikit/Tooltip';
+import { addAllFilesInCart, removeFilesFromCart } from '@ncigdc/dux/cart';
+import withFilters from '@ncigdc/utils/withFilters';
+
+const styles = {
+  dropdownContainer: {
+    left: 0,
+    right: 'auto',
+  },
+};
+
+const AddCaseFilesToCartButton = compose(
+  connect(state => ({ cart: state.cart })),
+  withState('isLoading', 'setIsLoading', false),
+  withFilters(),
+  withHandlers({
+    handleDropdownActivate: ({ relay, setIsLoading, filters, hasFiles }) => () => {
+      if (!hasFiles) {
+        // this case has no files to fetch.
+        return;
+      }
+      setIsLoading(true);
+      relay.setVariables({
+        isFileDataRequired: true,
+        filesFilters: filters,
+      }, (readyState) => _.some([readyState.ready, readyState.aborted, readyState.error]) && setIsLoading(false));
+    },
+  })
+)(({ hasFiles, files = [], filteredFiles = [], dispatch, handleDropdownActivate, isLoading }) => (
+  <Dropdown
+    dropdownStyle={styles.dropdownContainer}
+    dropdownClassName={isLoading ? 'hidden' : 'dropdown-menu'}
+    isDisabled={!hasFiles}
+    button={
+      hasFiles
+        ? (
+          <button className="btn btn-default dropdown-toggle fa fa-shopping-cart" style={{ padding: '0 4px' }}>
+            <span className={isLoading ? 'fa fa-spinner fa-spin' : 'caret'} style={{ marginLeft: '0.5rem' }} />
+            <span className="icon-btn-label">Case Actions</span>
+          </button>
+        )
+        : (
+          <Tooltip Component={<span>This case has no files.</span>}>
+            <button className="btn disabled fa fa-shopping-cart" style={{ padding: '0px 17px 0 4px' }} />
+          </Tooltip>
+        )
+    }
+    onActivate={handleDropdownActivate}
+  >
+    {
+      isLoading
+        ? <li> Loading case files... </li>
+        : ([
+          !!files.length && <li onClick={() => dispatch(addAllFilesInCart(files))} aria-label="Add all Case files to the Cart" role="button" key="addAll">
+            <ShoppingCartIcon style={{ marginRight: '1em' }} />Add all Case files to the Cart ({files.length})
+          </li>,
+          !!files.length && <li onClick={() => dispatch(removeFilesFromCart(files))} aria-label="Remove all Case files from the Cart" role="button" key="removeAll">
+            <TrashIcon style={{ marginRight: '1em' }} />Remove all Case files from the Cart ({files.length})
+          </li>,
+          !!filteredFiles.length && <li onClick={() => dispatch(addAllFilesInCart(filteredFiles))} aria-label="Add filtered Case files to the Cart" role="button" key="addFiltered">
+            <ShoppingCartIcon style={{ marginRight: '1em' }} />Add filtered Case files to the Cart ({filteredFiles.length})
+          </li>,
+          !!filteredFiles.length && <li onClick={() => dispatch(removeFilesFromCart(filteredFiles))} aria-label="Remove filtered Case files from the Cart" role="button" key="removeFiltered">
+            <TrashIcon style={{ marginRight: '1em' }} />Remove filtered Case files from the Cart ({filteredFiles.length})
+          </li>,
+        ])
+    }
+  </Dropdown>
+));
+
+export default AddCaseFilesToCartButton;
--- a/modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
+++ b/modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
@@ -23,11 +23,13 @@ const AddCaseFilesToCartButton = compose(
   connect(state => ({ cart: state.cart })),
   withState('isLoading', 'setIsLoading', false),
   withFilters(),
-  withProps(({ filters }) => ({
+  withProps(({ filters, cart, files = [], filteredFiles = [] }) => ({
     isFilteringOnFiles: filters && _.some(filters, (item) => {
       const content = item.hasOwnProperty('content') ? item.content : item;
-      return content.field.indexOf('files.') === 0;
+      return content && content.field && content.field.indexOf('files.') === 0;
     }),
+    filesInCart: _.intersectionBy(cart.files, files, 'file_id'),
+    filteredFilesInCart: _.intersectionBy(cart.files, filteredFiles, 'file_id'),
   })),
   withHandlers({
     handleDropdownActivate: ({ relay, setIsLoading, filters, isFilteringOnFiles, hasFiles }) => () => {
@@ -42,7 +44,7 @@ const AddCaseFilesToCartButton = compose(
       }, (readyState) => _.some([readyState.ready, readyState.aborted, readyState.error]) && setIsLoading(false));
     },
   })
-)(({ hasFiles, files = [], filteredFiles = [], dispatch, handleDropdownActivate, isLoading, isFilteringOnFiles }) => (
+)(({ hasFiles, files = [], filteredFiles = [], dispatch, handleDropdownActivate, isLoading, isFilteringOnFiles, filesInCart, filteredFilesInCart }) => (
   <Dropdown
     dropdownStyle={styles.dropdownContainer}
     dropdownClassName={isLoading ? 'hidden' : 'dropdown-menu'}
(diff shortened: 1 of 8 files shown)


modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
```diff
@@ -0,0 +1,87 @@
...
+    onActivate={handleDropdownActivate}
+  >
+    {
+      isLoading
+        ? <li> Loading case files... </li>
+        : ([
+          !!files.length && <li onClick={() => dispatch(addAllFilesInCart(files))} aria-label="Add all Case files to the Cart" role="button" key="addAll">
+            <ShoppingCartIcon style={{ marginRight: '1em' }} />Add all Case files to the Cart ({files.length})
+          </li>,
+          !!files.length && <li onClick={() => dispatch(removeFilesFromCart(files))} aria-label="Remove all Case files from the Cart" role="button" key="removeAll">
```
SidSyrus: I think you can use `toggleFilesInCart` function to remove files from cart. Don't really need to create a new function.

modules/node_modules/@ncigdc/uikit/Dropdown.js
```diff
@@ -35,7 +37,7 @@ const Dropdown = ({
         onMouseUp={mouseUpHandler}
         onClick={() => setTimeout(() => setActive(false))}
       >
-        {Children.map(children, child =>
+        {Children.map(_.filter(children, React.isValidElement), child =>
```
SidSyrus: You could use the native `filter` method here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

modules/node_modules/@ncigdc/uikit/withToggle.js
```diff
@@ -1,20 +1,42 @@
...
   return class Toggle extends Component {
+    static propTypes = {
+      // Allow active state changes to trigger actions such as toggling relay variables
+      onActivate: PropTypes.func,
+      onDeactivate: PropTypes.func,
+      active: PropTypes.bool,
+    };
+
+    static defaultProps = {
+      onActivate: _.noop,
```
cy: benefits over `() => {}` and then not depending on lodash?

modules/node_modules/@ncigdc/uikit/withToggle.js
```diff
@@ -1,20 +1,42 @@
...
+    };
+
+    toggleActive = (): void => this.setActive(!this.state.active);
 
     render() {
       return (
         <ComposedComponent
           {...this.state}
-          {...this.props}
+          {..._.omit(this.props, _.keys(Toggle.defaultProps))}
```
cy: same question, why not `Object.keys()`

modules/node_modules/@ncigdc/containers/CaseTr.js
```diff
@@ -35,7 +36,17 @@ export type TProps = {|
   theme: Object,
 |};
 
-export const CaseTrComponent = ({ node, index, theme }: TProps) => {
+function massageRelayFiles(files: Array): Array {
+  return _.get(files, 'hits.edges', [])
```
cy: lol @ the function name. lodash-less `(files.hits || { edges : [] }).edges.map` instead? Oh and `files` is typed Array in the signature, should it be an Object?

modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
cheapsteak: @SidSyrus  `toggleFilesInCart` seems like it'd be potentially buggy if used to perform add or remove operations on an array of files in a cart

What happens here:

Case A has files [1, 2, 3]
Case B has files [3, 4, 5]

If I click on "add case files" for Case A, then "remove case files" for Case B
I _want_ to end up with `[1,2]`, but would `toggleFilesInCart` not result in `[1,2,4,5]` (undesired)?

modules/node_modules/@ncigdc/uikit/Dropdown.js
cheapsteak: Appreciate the tip but this is intentional  

Using the native filter naievely would result in `Cannot read property 'filter' of undefined` when there are no children (children is undefined), or when there is a single string/number/react element.

The code becomes rather unnecessarily verbose, for the benefit of not using a function in a library we're using anyway. Cost/benefit doesn't work out

modules/node_modules/@ncigdc/uikit/withToggle.js
cheapsteak: I don't have strong preferences on this one. Academic memory and initialization benefits of sharing a single instance of noop instead of creating separate functions, but that problably comes down to less than a few kb and a few ms

modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
SidSyrus: I don't think Cases share files. If you look at the Production Portal you will see it won't show remove files option if the files for that particular case is not in the cart.

modules/node_modules/@ncigdc/uikit/withToggle.js
cheapsteak: I can definitely prefer native methods if that's desired. Figured I'm using `_.omit` anyway, didn't see the benefit of using `Object.keys` here. Let's add that to our eslintrc if we prefer native methods over lodash ones

JDD thinks there's a benfit of preferring lodash functions, but then of course he would :)
https://github.com/eslint/eslint/issues/5017#issuecomment-173607411

modules/node_modules/@ncigdc/uikit/Dropdown.js
SidSyrus: How come `Children.map` won't throw an error?

modules/node_modules/@ncigdc/uikit/withToggle.js
cy: I've sometimes run into strangeness with the build tool attempting to call lodash2 when lodash4 was in package.json and then failing, especially when running jest tests. That's why I'd prefer to not use lodash if it's simple enough not to.

modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
cheapsteak: I don't think it's a good idea to use an incorrect implementation just because it happens not to break under current assumptions that aren't guaranteed to always be true  

"Remove case files from cart" should "Remove case files from cart", not "remove files from cart, unless files weren't already in the cart, in which case it'll unexpectedly add the files to the cart"

modules/node_modules/@ncigdc/uikit/withToggle.js
cheapsteak: `() => {}` it is 👍

modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
SidSyrus: I don't think its an incorrect implementation. In your scenario you are removing `3` which also exist in Case A but the user only asked to remove files from Case B. So shouldn't the result be [1,2,3], only files for Case A? And I was suggesting using `toggleFilesInCart` only to remove files.

modules/node_modules/@ncigdc/containers/CaseTr.js
cheapsteak: >files is typed Array in the signature, should it be an Object

Ack, yup you're right. Actually tripped myself up a bit on that at first and forgot to change the signature

>  (files.hits || { edges : [] }).edges.map
Looks like it'd have to be `((files || {}).hits || { edges: [] }).edges` since `files` is sometimes passed in as `undefined`. That looks a bit harder to parse than `_.get(files, 'hits.edges', [])` imo

Don't mind changing that to lodashless if you feel strongly about it, but would much prefer to take a look at our jest setup to see if we can fix the lodash2 issue there :p  What do you think?

cy: Well that style of null checking is used all over this repo already instead of _.get(), so I'm pretty used to not using _.get() haha. But I've already banged my head over the jest setup way too much so if you wanna help in this PR https://github.com/NCI-GDC/portal-ui/pull/1268, if _.uniq is called it'll try to call lodash2 instead of 4 and die.

modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
cheapsteak: Apologies let me rephrase/simplify that

With a call to 
`toggleFilesInCart([1, 2, 3])` 
Whether [1, 2, 3] is in the cart after the call depends on whether [1, 2, 3] was the cart before the call. Places that use it would also have to guard against multiple invocations (the second time that function gets called it becomes an "add" instead of a "remove")

With
`removeFilesFromCart([1, 2, 3])` 
I have the guarantee that after that function call, files [1, 2, 3] will not be in the cart, regardless of whether they were in the cart, regardless of how many times that function is called

I don't actually see the value of having a "toggle" with an xor with an array of files. Are we currently using `toggleFilesInCart` when we actually mean to "remove"? If we want to DRY up code, perhaps we can refactor instances where we're using `toggle` and replace it with a more explicit `remove`. What do you think?

modules/node_modules/@ncigdc/uikit/Dropdown.js
cheapsteak: That function is quite robust and does a lot under the hood :)

https://github.com/facebook/react/blob/6810627a91f2020caee7fecf21580187dc6ef0fc/src/isomorphic/children/ReactChildren.js#L161-L168

https://github.com/facebook/react/blob/6810627a91f2020caee7fecf21580187dc6ef0fc/src/shared/utils/traverseAllChildren.js#L98-L192

modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
SidSyrus: As I mentioned earlier I am suggesting using `toggleFilesInCart` only to remove files. 

We know which button is for **Add all** and which button is for **Remove all**. You are already using `addAllFilesInCart` to add, which is perfectly fine. Now for **Remove all** if we use `toggleFilesInCart` and pass the files, the function will know that its task now is to remove them not to add. Therefore it works perfectly fine. The logic for `toggleFilesInCart` and `removeFilesFromCart` are almost similar. 

I agree the name doesn't justify the action. If you insist on using `removeFilesFromCart` function I would suggest taking a look at it as it would need some update. We could use `toggleFilesInCart` only for singular file and `removeFilesFromCart` for multiple files.

modules/node_modules/@ncigdc/uikit/Dropdown.js
SidSyrus: That is not the native `Array.map()` ?

alex-wilmer: Nope. You can tell by the function signature. Array#map takes a callback as the first argument

https://facebook.github.io/react/docs/react-api.html#react.children

modules/node_modules/@ncigdc/components/AddCaseFilesToCartButton.js
cheapsteak: >As I mentioned earlier I am suggesting using `toggleFilesInCart` only to remove files.

I understand, that's where I'm saing the issue lies. Let me try to rephrase

When I'm trying to remove files from a cart, with a call to 
`toggleFilesInCart([1, 2, 3])` 
I have no guarantee that the cart will no longer contain the files I passed in.
Whether [1, 2, 3] is in the cart after the call depends on whether [1, 2, 3] was in the cart before the call. If a file is passed in to the function that wasn't already in the cart, it would be added to the cart. Places that use the function would also have to guard against multiple invocations. The second time that function gets called it becomes an "add" instead of a "remove"

This kind of unexpected behavior of  "wanting to only remove files" but the result of the call to `toggleFilesInCart` actually adding files, is why the `removeFilesFromCart` function was added - so that there is a guarantee that when there is a desire to remove files, under no circumstances would files be added to the cart.

The defining logic between `toggleFilesInCart` and `removeFilesFromCart` is that one is an "xor", the other is a "difference". Sometimes the xor results in a difference. Other times it results in an addition. (places that are currently using `addAllFilesInCart` (which should perhaps be renamed to `addFilesToCart`) could also be changed to use`toggleFilesInCart` with similar drawbacks)

The only instance where an xor on the cart by an array of passed in files would be useful would be if we desire to have a function that we call, passing in an array of files, adding the ones that don't exist and removing the ones that do. 

```
cart = [1, 2, 3];

toggleFilesInCart([1, 2, 3, 4, 5, 6]);

// cart is now [4, 5, 6];
```

cheapsteak:
@cy @SidSyrus ready for re-review

Removed instances of _.noop to use `() => {}`, and `_.keys` to `Object.keys`

I'm hesitant to refactor instances of `toggleFilesInCart` to use `removeFilesFromCart` in this PR. Perhaps once the cart settles and all the PRs that might be depending on `toggleFilesInCart`'s existence is merged

cy:
All right so not really following the message thread above, kind of long. But `Remove from Cart (39)` when cart is empty still gives `30 files removed from Cart` message, isn't that strange at least?

cheapsteak:
That I can fix :)

modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -92,6 +92,28 @@ function toggleFilesInCart(incomingFile) {
...
+function removeFilesFromCart(files) {
+  return (dispatch, getState) => {
+    const filesToRemove = _.isArray(files) ? files : [files];
+    const existingFiles = getState().cart.files;
+    const nextFiles = _.differenceBy(existingFiles, filesToRemove, 'file_id');
+    dispatch(notify(getNotificationComponent('remove', `remove/${filesToRemove.file_name}`, {
+      action: 'Removed',
+      file: filesToRemove.file_name || filesToRemove.length,
+      fileText: filesToRemove.file_name ? '' : 'files',
+      prepositon: 'from',
```
cy: screen shot 2017-02-22 at 5 01 07 pm
missing a space?

cheapsteak:
@cy updated PR, now showing numbers in notifications and dropdown options. Added a space

modules/node_modules/@ncigdc/uikit/Dropdown.js
SidSyrus: aah I see!

cy:
cool, cool - can you check the AddToCartSingle buttons on the files tab and the files entity page? Seems broke.

modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -92,6 +92,29 @@ function toggleFilesInCart(incomingFile) {
   };
 }
 
+function removeFilesFromCart(files) {
+  return (dispatch, getState) => {
+    const filesToRemove = _.isArray(files) ? files : [files];
+    const existingFiles = getState().cart.files;
+    const nextFiles = _.differenceBy(existingFiles, filesToRemove, 'file_id');
+    const filesRemoved = existingFiles.length - nextFiles.length;
+    dispatch(notify(getNotificationComponent('remove', `remove/${filesToRemove.file_name}`, {
```
SidSyrus: Wouldn't `filesToRemove.file_name` be undefined if its an array? 
Also, I noticed the `id` we pass to notification has to be unique. In this case if you remove files once wait for notification to go away and remove again, the new notification won't show up. This also happens with Files table. Somehow we need to generate unique IDs each time. Do you mind taking a look? Or should we create a different ticket for this as its a bug with Files table as well?

cheapsteak: Nice catch

I copied and pasted that part from [toggleFilesInCart](https://github.com/NCI-GDC/portal-ui/blob/baae998a578fba75704da6ddec410dda5c9a9793/modules/node_modules/%40ncigdc/dux/cart.js#L66-L77). How is it handled there?

cheapsteak: For the unique ids, I think we could generate that in the `getNotificationComponent`. I'll check with the others to see if that'd ok

SidSyrus: It works out for `toggleFilesInCart` because its being used for singular file so something like `filesToRemove.file_name` will always be unique. However for `addAllFilesInCart` or `removeFilesFromCart` where files could be in array the same logic doesn't work out. So either we would have to generate unique IDs or update notification to show it every time despite of what the ID is. This is where I think its being checked. Not totally sure though.

https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/uikit/Notification.js#L101

cheapsteak: Thanks
Switched to using timestamp in the id
https://github.com/NCI-GDC/portal-ui/pull/1277/commits/490701336d00b4dc02cf6246797d94d209ace25e

modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -98,7 +98,7 @@ function removeFilesFromCart(files) {
     const existingFiles = getState().cart.files;
     const nextFiles = _.differenceBy(existingFiles, filesToRemove, 'file_id');
     const filesRemoved = existingFiles.length - nextFiles.length;
-    dispatch(notify(getNotificationComponent('remove', `remove/${filesToRemove.file_name}`, {
+    dispatch(notify(getNotificationComponent('remove', `remove/${(new Date()).getTime()}`, {
```
SidSyrus: Could you do this for `addAllFilesInCart` as well as `removeAllInCart`?

cheapsteak:
@cy I think those are broken on the main `relay` branch as well

modules/node_modules/@ncigdc/dux/cart.js
cheapsteak: Done

cheapsteak:
@cy I think the single files breaking is related to inner hits
Should that be on a separate ticket?

cy:
yeah, also see that on the relay branch sure.

Am I imagining things or did the notification used to show the file name when it's one file?
screen shot 2017-02-23 at 3 42 45 pm
if not that can be a separate ticket too
👍

cheapsteak:
It's always been like that for adding case files :)

Would prefer to move that to a separate ticket, I have an idea to take that notification box to another level :D

Currently the behaviour for whether the users sees "Added/removed  N files" or "Added/removed FILENAME" depends on whether the _type_ of the argument passed in to the add-to/remove-from cart function was an array or a single object.

```
addAllFilesInCart(file)
// -> "Added filename.filetype to cart"

addAllFilesInCart([file1])
// -> "Added 1 files to cart"

addAllFilesInCart([file1, file2, file3, file4])
// -> "Added 4 files to cart"
```

I think the UX may be improved with this

```
addAllFilesInCart(file)
// -> "Added file <em>filename.filetype</em> to cart"

addAllFilesInCart([file])
// -> "Added file <em>filename.filetype</em> to cart"

addAllFilesInCart([file1, file2, file3, file4])
// -> "Added <em>3</em> files <em>filename1.filetype, filename2.filetype, filena ...</em> to cart"
```

The max length of the filenames to keep could be played around with and tweaked (I'm thinking maybe around 250 characters?)

The benefits would be more consistent messaging, and also provide a sort of visual signature to help differentiate between notifications (e.g. when adding files for cases that have the same number of files)

cy:
You're saying to also show the filenames when addAllFilesInCart() is passed an array but limit the word count? Fine by me!

cheapsteak:
Yup!
Is this good to merge for now? 
Would bundle that messaging change with refactoring of toggleFilesInCart (which should probably internally reuse addAllFilesInCart (which would be renamed to addFilesToCart) and removeFilesFromCart), but want to wait for dust to settle on PRs touching the cart

cy:
👍

PR #1278, opened 2017-02-21
🐛 [PRTL-627]  Biospecimen details section update

SidSyrus:
- GraphQL query update to get fields mentioned in spec
- Passing type of entity when it is selected


SidSyrus:
@Jephuff Ready for review again.

PR #1289, opened 2017-02-24
✨ add link on cases column

Jephuff:


modules/node_modules/@ncigdc/containers/FileTable.js
```diff
@@ -139,6 +139,7 @@ export const FileTableQuery = {
                 total
```
shanewilson: I would only ask for 1 case

shanewilson:
I don't think we need the approved tag - since there are reviews built in now

modules/node_modules/@ncigdc/containers/FileTable.js
shanewilson: I would only ask for 1 case since you only need the actual data from the first hit

Jephuff: It's using the full array in at least one spot 
https://github.com/NCI-GDC/portal-ui/blob/a50bfe6d3069373d5da9a1165979abaada7318a5/modules/node_modules/%40ncigdc/containers/FileTr.js#L72

Jephuff: makes sense, updated 👍

PR #1290, opened 2017-02-24
✨ refactor analysis calls, and giant top level relay queries into meaningful containers

Jephuff:


alex-wilmer:
that has been split into multiple PRS

PR #1292, opened 2017-02-24
✨ [PRTL-703] custom facets modal : allow filtering out facets that are not useful

cheapsteak:
@shanewilson You were right, it did turn out to be simpler to reuse stuff from CaseAggreagations and FileAggregations :)

### DIFF
--- a/modules/node_modules/@ncigdc/components/Aggregations/TermAggregation.js
+++ b/modules/node_modules/@ncigdc/components/Aggregations/TermAggregation.js
@@ -2,7 +2,7 @@
 
 import React from 'react';
 import LocationSubscriber from '@ncigdc/components/LocationSubscriber';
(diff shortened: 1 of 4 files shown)


cheapsteak:
@cy great ideas
Implemented and pushed!

cy:
doesn't seem to happen on the relay branch, but flipping between the 'Case' and 'File' facet tabs causes some of the custom facets to all become 'No data for this field' when flipping back.
screen shot 2017-03-01 at 3 16 42 pm
screen shot 2017-03-01 at 3 16 55 pm

cheapsteak:
@cy great catch, fixed up TermAggregations to reactively determine what its filtered buckets are. It used to set the value in the instant of `componentDidMount` and when a keyword is clicked, now it computes each time `buckets` or `filters` changes.

ready for re review

cheapsteak:
Sounds good!

PR #1296, opened 2017-02-27
🐛 [PRTL-685] link the Available File per Data Category counts to the Data Repository 

SidSyrus:
- Link updated to use repository page 
- Passing filter through JSURL stringify function

### DIFF
--- a/modules/node_modules/@ncigdc/components/MostAffectedCases.js
+++ b/modules/node_modules/@ncigdc/components/MostAffectedCases.js
@@ -3,6 +3,8 @@
 import React from 'react';
 import withSize from 'react-sizeme';
 import { compose } from 'recompose';
+import JSURL from 'jsurl';
+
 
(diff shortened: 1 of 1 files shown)


modules/node_modules/@ncigdc/components/MostAffectedCases.js
```diff
@@ -125,10 +127,10 @@ const MostAffectedCases = ({
                   dataCategorySummary[DATA_CATEGORIES[k].full] ? (
                     <a
                       href={
-                        `/search/f?filters=${makeFilter([
+                        `/repository?filters=${JSURL.stringify(makeFilter([
```
cy: if you use the `RepositoryFileLink` component instead, wouldn't need to manually call JSURL.stringify example https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/containers/CaseTr.js#L53

SidSyrus:
@cy ready for review again.

PR #1298, opened 2017-02-27
🐛PRTL-690 filter most frequent mutations by project

cheapsteak:
Project Entity page
Most Frequent Mutations and Most Frequently Mutated Genes now accounting for project id when making request

Case Entity Page
Most Frequent Mutations accounting for case id when making request

### DIFF
--- a/modules/node_modules/@ncigdc/components/Case.js
+++ b/modules/node_modules/@ncigdc/components/Case.js
@@ -8,7 +8,7 @@ import { connect } from 'react-redux';
 
 import LocationSubscriber from '@ncigdc/components/LocationSubscriber';
 import withRouter from '@ncigdc/utils/withRouter';
-import { makeFilter, mergeQuery } from '@ncigdc/utils/filters';
+import { makeFilter, mergeQuery, setFilter, replaceFilters } from '@ncigdc/utils/filters';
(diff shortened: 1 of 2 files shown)


cheapsteak:
@Jephuff care to review this? touches mostly your code (but very minimal amount of code touch)

modules/node_modules/@ncigdc/components/Case.js
```diff
@@ -281,13 +282,14 @@ const Case = ({
         <LocationSubscriber>{(ctx: {| pathname: string, query: TRawQuery |}) => {
           const { filters } = ctx.query || {};
           const currentFilters = parseFilterParam(filters, { op: 'and', content: [] });
+          const componentFilters = _.mergeWith(currentFilters, setFilter({ field: 'cases.case_id', value: p.case_id }), (a, b) => (Array.isArray(a) ? a.concat(b) : undefined));
```
Jephuff: Might be good to using the methods in the filters utils. something like this https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/components/Oncogrid/getQueries.js#L19

cheapsteak:
Thanks for the usage example! That looks so much better @Jephuff ready for re-review?

cheapsteak:
@Jephuff Thank you!

PR #1300, opened 2017-02-28
✨ [PRTL-718] underlines for table headers with tooltips

Jephuff:


Jephuff:
this is ready for review

PR #1303, opened 2017-02-28
🐛 [PRTL-712] ensure valid values and update tooltips

Jephuff:


Jephuff:
tooltips will not update until this is merged https://github.com/oncojs/oncogrid/pull/30

shanewilson:
does that branch need to be merged before this one? or just you won't see the update until then?

PR #1306, opened 2017-03-01
♻️ [PRTL-698] use dropdown for download buttons

Jephuff:


Jephuff:
Needs to be tested with lolliplot working to ensure it downloads.

Jephuff:
tested

PR #1308, opened 2017-03-01
✨ [PRTL-738] scroll to top on section change

Jephuff:


modules/node_modules/@ncigdc/components/PortalContainer.js
```diff
@@ -27,12 +28,26 @@ import ProgressContainer from '@ncigdc/components/ProgressContainer';
...
 
 const SkipLink = styled.a({ position: 'absolute', left: '-999px', backgroundColor: 'black', zIndex: 1000 });
 
+const enhance = compose(
+  connect(store => ({ notifications: store.bannerNotification })),
+  withRouter,
+  lifecycle({
+    componentDidMount(): void {
+      // eslint-disable-next-line fp/no-this, fp/no-mutation
+      this.removeListen = this.props.listen(() => window.scrollTo(0, 0));
```
shanewilson: when does this trigger?

Jephuff: whenever any route change happens. I didn't think about query parameters. I'll fix that :)

Jephuff: should be good now.

Jephuff:
This is ready for review again

cheapsteak:
This isn't showing up on deploybot, any way we can make it show up?

Jephuff:
force pushed and it deployed, but can't seem to get pages to load consistently. 
to test go pages project TCGA-CHOL -> gene PBRM1 -> projects -> TCGA-CHOL
then scroll down and click `PBRM1` again, because the gene was already loaded, it would have been scrolled down the gene page.

PR #1309, opened 2017-03-01
:bug: PRTL-717, PRTL-684 fix BarChart links

cy:
- change href prop on BarChart to onClick handler instead


cy:
@Jephuff yup - done

Jephuff:
ooh, could we also add `cursor: pointer`?

cy:
good point, thanks for pointing it out, pointer added. 😛

PR #1311, opened 2017-03-02
Prtl 598 table sorting

Jephuff:
requires https://github.com/NCI-GDC/gdcapi/pull/706.

### DIFF
--- a/modules/node_modules/@ncigdc/routes/AnnotationsRoute.js
+++ b/modules/node_modules/@ncigdc/routes/AnnotationsRoute.js
@@ -8,7 +8,7 @@ import { parse } from 'query-string';
 
 import { handleStateChange } from '@ncigdc/dux/relayProgress';
 import AnnotationsPage from '@ncigdc/containers/AnnotationsPage';
-import { parseIntParam, parseFilterParam } from '@ncigdc/utils/uri';
+import { parseIntParam, parseFilterParam, parseJSURLParam } from '@ncigdc/utils/uri';
 
 import { viewerQuery } from './queries';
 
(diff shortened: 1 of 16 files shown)


modules/node_modules/@ncigdc/routes/AnnotationsRoute.js
```diff
@@ -5,25 +5,30 @@ import React from 'react';
...
   static prepareParams = ({ location: { search } }) => {
     const q = parse(search);
+    const annotationsSort = q.annotations_sort ? JSURL.parse(q.annotations_sort) : DEFAULT_ANNOTATIONS_SORT;
+
     return ({
       offset: parseIntParam(q.offset, 0),
       size: parseIntParam(q.size, 20),
       filters: parseFilterParam(q.filters, null),
-      annotations_sort: q.annotations_sort || null,
+      annotations_sort: annotationsSort,
```
shanewilson: `parseFilterParam(q.sort, null)`?

parseFilter just runs it through `jsurl.parse`

shanewilson:
the sort api works? makes sense to use?

Jephuff:
Yep, I haven't done anything crazy, just passing an array of objects like `[{ field: 'primary_site', order: 'asc' }]` and it looks like it's working. 
I did need to add the `.raw` but I think you warned me about that and it's easy enough to change in the future if the requirement changes.

modules/node_modules/@ncigdc/routes/AnnotationsRoute.js
Jephuff: ahh makes sense, maybe I should change the name to something like `parseJSURLParam`?

PR #1315, opened 2017-03-03
🐛 [PRTL-769] Survival plot legend change

SidSyrus:
Legend change. 
Link to mutation entity page from bar chart updated


Jephuff:
actually, get you get rid of the frequentMutation.js change, @cy fixes that in this pull request https://github.com/NCI-GDC/portal-ui/pull/1309

PR #1325, opened 2017-03-06
:pencil: PRTL-773 update table header copy

cy:
```
Reported by Vincent.
Please update labels in table header.
In Most Frequent Mutations tables:
>>"# Affected Cases in [Project_id] Case" >> "# Affected Cases in [Project_id]" (remove Case word)
In Most Frequently Mutated Genes and Most Frequent Mutations tables:
"# Affected Cases Across all Projects" >> "# Affected Cases in all Projects" (remove Across word)
```

### DIFF
--- a/modules/node_modules/@ncigdc/components/MutationTable.js
+++ b/modules/node_modules/@ncigdc/components/MutationTable.js
@@ -109,7 +109,7 @@ const MutationTable: TMutationTable = ({
             }
             style={tableToolTipHint()}
           >
-            # Affected Cases<br />in {projectId} Case
+            # Affected Cases<br />in {projectId}
           </Tooltip>
         ),
       }] : []),


SidSyrus:
The ticket also mentions 'Most Frequently Mutated Genes' table. I think its in [`GeneTable.js`](https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/@ncigdc/components/GeneTable.js#L56) file.

cy:
@SidSyrus thanks for the catch, found it

modules/node_modules/@ncigdc/components/MutationTable.js
```diff
@@ -118,7 +118,7 @@ const MutationTable: TMutationTable = ({
               </span>
             }
           >
-            # Affected Cases<br /> Across all Projects
+            # Affected Cases<br /> all Projects
```
SidSyrus: missing `in`

alex-wilmer: or maybe capitalize the `A` in all so its less of a sentence.

```
# Affected Cases
All Projects
```

cy: maybe both...? urgh grammar

SidSyrus: Well the ticket says
`# Affected Cases Across all Projects >> # Affected Cases in all Projects`

cy: yeah I forgot the `in`. But thinking about it, `all` is not a preposition, article or conjunction in titles. It looks like a title because all the other words are capitalized.
screen shot 2017-03-06 at 12 22 13 pm
:woman_shrugging:

alex-wilmer: Try not think about it as a sentence (it's table header after all)

It would be best as `# Affected Cases - All Projects` but the line break could also take the place of the hyphen

cy: `# Affected Cases (All Projects)`

alex-wilmer: Yea that works!

cy: @phuongmy ok with `# Affected Cases (All Projects)`?

cy:
Fixed conflict & @phuongmy updated the ticket to be just removing the "Case" word.

PR #1330, opened 2017-03-06
[PRTL-757] Truncate Mutation ID to 8 characters 

SidSyrus:
Truncating Mutation ID to 8 characters to the places mentioned in [specs](https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=GDCSPECS&title=Plan+B+-+Use+shortened+Mutation+ID%2C+pending+short+SSM+ID+availability)

### DIFF
--- a/modules/node_modules/@ncigdc/utils/survivalplot.js
+++ b/modules/node_modules/@ncigdc/utils/survivalplot.js
@@ -64,7 +64,7 @@ export const getSurvivalCurves = async ({ value, field, slug, currentFilters }: 
         key: `${slug || value}-mutated`,
         value: (
           <span>
-            S<sub>1</sub> (N={rawData.results[0].donors.length}) - <Symbol>{slug || value}</Symbol> Mutated Cases
+            S<sub>1</sub> (N = {rawData.results[0].donors.length}) - <Symbol>{slug || value}</Symbol> Mutated Cases
           </span>
         ),
       },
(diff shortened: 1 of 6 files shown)


shanewilson:
i don't think we should add `...` or tooltips. Just pretend the id is only 8 chars.

modules/node_modules/@ncigdc/utils/survivalplot.js
```diff
@@ -41,7 +41,7 @@ export const getDefaultCurve = async ({ slug, totalCases, currentFilters }: TPro
   };
 };
 
-export const getSurvivalCurves = async ({ value, field, slug, currentFilters }: TPropsMulti): Promise<Object> => {
+export const getSurvivalCurves = async ({ value, field, slug, currentFilters, shortMutationId }: TPropsMulti): Promise<Object> => {
```
Jephuff: I don't think there is any need to add this logic here seeing as it's only temporary. just pass the 8 char string as `slug`

SidSyrus: aah true

modules/node_modules/@ncigdc/utils/survivalplot.js
```diff
@@ -64,15 +64,15 @@ export const getSurvivalCurves = async ({ value, field, slug, currentFilters }:
         key: `${slug || value}-mutated`,
         value: (
           <span>
-            S<sub>1</sub> (N = {rawData.results[0].donors.length}) - <Symbol>{slug || value}</Symbol> Mutated Cases
+            S<sub>1</sub> (N={rawData.results[0].donors.length}) - <Symbol>{slug || value}</Symbol> Mutated Cases
```
Jephuff: Did you mean to do this? I think it's undoing something Christine did.

PR #1332, opened 2017-03-06
:bug: [PRTL-745] Case table - # Files is incorrect

SidSyrus:
Added missing data category.


shanewilson:
case has a `file_count` directly under summary - can we use this number for the file count column?

SidSyrus:
Oh! Ok I will use that then. Would you still like to have that category added to the mapping?

shanewilson:
we can - not sure it matters though since I don't think we show that column on the UI do we?

SidSyrus:
Not that I know of.

PR #1336, opened 2017-03-06
Prtl 777 oncogrid legend

Jephuff:


Jephuff:
Ahh, yep, that shouldn't have been out of 100 :) 
fixed it now.

PR #1339, opened 2017-03-07
♻️✨ FrequentMutations Relay containers

alex-wilmer:


modules/node_modules/@ncigdc/containers/FrequentMutationsTable.js
```diff
@@ -0,0 +1,364 @@
...
+  }
+  const frequentMutations = mapData(ssms.hits.edges.map(x => x.node), projectId);
+  return (
+    <div>
+      <Row style={{ backgroundColor: 'white', padding: '1rem', justifyContent: 'space-between' }}>
+        <Showing
+          docType="mutations"
+          prefix={'fmTable'}
+          params={relay.route.params}
+          total={ssms.hits.total}
```
Jephuff: A little bit of weirdness with this 
image
I think it just needs to use `parseInt` for the size/offset inside Showing.

Jephuff:
We should make sure these changes get made here too :) 
https://github.com/NCI-GDC/portal-ui/pull/1330
https://github.com/NCI-GDC/portal-ui/pull/1341

cy:
if you merge this I can redo #1341

alex-wilmer:
@cy it's okay I put it in the table for now. if we ever have the chart actually on its own we can deal with that then.

cy:
oki, thought this looked pretty weird though
screen shot 2017-03-08 at 3 18 49 pm
so reusing #1341 to hide the chart

alex-wilmer:
ooopps maybe just `if (!ssms.hits.length) return null`?

PR #1340, opened 2017-03-07
PRTL-794 Case page 4th number issue - Most Frequent Mutations table

SidSyrus:
Calculating totalNumCases for case page

### DIFF
--- a/modules/node_modules/@ncigdc/components/Case.js
+++ b/modules/node_modules/@ncigdc/components/Case.js
@@ -53,7 +53,6 @@ const Case = ({
   node,
   theme,
   numCasesAggByProject,
-  totalNumCases = Object.keys(numCasesAggByProject).reduce((sum, b) => sum + numCasesAggByProject[b], 0),
   push,
   query,
   dispatch,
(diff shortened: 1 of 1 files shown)


modules/node_modules/@ncigdc/components/Case.js
```diff
@@ -51,9 +51,9 @@ const enhance = compose(
 
 const Case = ({
   node,
-  totalNumCases,
   theme,
   numCasesAggByProject,
+  totalNumCases = Object.keys(numCasesAggByProject).reduce((sum, b) => sum + numCasesAggByProject[b], 0),
```
Jephuff: Seeing as this isn't going to be passed in, might as well take it out of the props.

SidSyrus: What if we need to pass that in later on? That way this would be the default value and if we do pass in the future, it will override it.

shanewilson: we can always update the code if that happens. I think it is preferable to code to known specs instead of trying to guess the future. it keeps things simple as long as possible

SidSyrus:
Not needed any more. As [6d69722](https://github.com/NCI-GDC/portal-ui/commit/6d697226c2074fa6ac626d0e913789047f5c9da6) takes care of it.

PR #1341, opened 2017-03-07
:bug: PRTL-795 MostFrequentMutationsContainer display none message

cy:
~move the 'No mutations data found' message from `FrequentMutations` to `FrequentMutationsContainer` because it was never being reached if there were none.~
now hiding the chart when no hits. also updated the barchart herfs to onClicks


alex-wilmer:
👍

PR #1349, opened 2017-03-08
🐛 [PRTL-793] fix lolliplot transcript selection

Jephuff:


alex-wilmer:
Yeah cool this works 👍 can merge this and I'll sort out the relay container stuff

PR #1352, opened 2017-03-09
:bug: [PRTL-799] align bar chart left

SidSyrus:
- Making Bar charts half the width and aligning them left side

### DIFF
--- a/modules/node_modules/@ncigdc/components/ProjectVisualizations.js
+++ b/modules/node_modules/@ncigdc/components/ProjectVisualizations.js
@@ -155,7 +155,7 @@ const ProjectVisualizations = enhance(({
                 numCasesAggByProject={numCasesAggByProject}
                 defaultFilters={fmFilters}
                 cohort={viewer.frequentMutationsChartFragment}
-                showSurvivalPlot={true}
+                showSurvivalPlot
               />
             </Column>
             <Column flex="1">
--- a/modules/node_modules/@ncigdc/components/ProjectVisualizations.js
+++ b/modules/node_modules/@ncigdc/components/ProjectVisualizations.js
@@ -155,6 +155,7 @@ const ProjectVisualizations = enhance(({
                 numCasesAggByProject={numCasesAggByProject}
                 defaultFilters={fmFilters}
                 cohort={viewer.frequentMutationsChartFragment}
+                showSurvivalPlot={true}
               />
             </Column>
             <Column flex="1">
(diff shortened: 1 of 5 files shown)


modules/node_modules/@ncigdc/components/ProjectVisualizations.js
```diff
@@ -155,6 +155,7 @@ const ProjectVisualizations = enhance(({
                 numCasesAggByProject={numCasesAggByProject}
                 defaultFilters={fmFilters}
                 cohort={viewer.frequentMutationsChartFragment}
+                showSurvivalPlot={true}
```
Jephuff: linter doesn't like this, you can remove `={true}`

modules/node_modules/@ncigdc/containers/FrequentMutationsChart.js
```diff
@@ -48,7 +48,7 @@ const FrequentMutationsChartComponent = compose(
 
   const bandWidth = (
     (width - chartMargin.right - chartMargin.left) /
-    (ssms.hits.edges.length + 1) / (showSurvivalPlot ? 2 : 1)
+    (ssms.hits.edges.length + 1) / (showSurvivalPlot ? 1 : 2)
```
Jephuff: Shouldn't this just always be divided by 2?

modules/node_modules/@ncigdc/containers/FrequentMutationsChart.js
```diff
@@ -63,8 +63,15 @@ const FrequentMutationsChartComponent = compose(
             tooltipHTML="Download image or data"
           />
         </div>
-        <div style={graphTitle()}>Distribution of Most Frequent Mutations</div>
-        <Row id="mutation-chart" style={{ padding: '10px', justifyContent: 'center' }}>
+        <div
+          style={{
+            ...graphTitle(),
+            width: showSurvivalPlot ? width : ((width / 2) - chartMargin.left - chartMargin.right),
```
Jephuff: this is a bit strange, and also leaves the download button way over to right.
It might be better to have a parent container for the barchart/button/title that either has it's width set or is `display: inline-block`

modules/node_modules/@ncigdc/containers/FrequentMutationsChart.js
SidSyrus: Not in this particular case.

Jephuff: Ahh, I see it now. 
If thats the case, then it probably makes sense to just make the graph always full width, and then the parent can size the container.

SidSyrus: >  make the graph always full width, and then the parent can size the container

Not sure if I follow.

Jephuff: so, never divide by two, just use `(ssms.hits.edges.length + 1)`, then where `FrequentMutationsChart` is used we just make it the size we actually want, so in cases where there isn't a survivalplot we can just add width 50%

PR #1354, opened 2017-03-09
🐛💄 [PRTL-800 & 801] pagination selected page style change

SidSyrus:
- Style was overwritten by another class so made the CSS rule `!important`

### DIFF
--- a/modules/node_modules/@ncigdc/components/Pagination/Sizes.js
+++ b/modules/node_modules/@ncigdc/components/Pagination/Sizes.js
@@ -12,7 +12,7 @@ export type TProps = {|
   sizes?: Array<number>,
 |};
 
-const Sizes = ({ sizes = [20, 40, 60, 80, 100], size, prfSize }: TProps) => {
+const Sizes = ({ sizes = [10, 20, 40, 60, 80, 100], size, prfSize }: TProps) => {
   return (
     <span>
       <Dropdown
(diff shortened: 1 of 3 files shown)


shanewilson:
what is taking priority?

bessierogel:
At the time these are UI tickets that had been raised by Vincent and is deemed part of priority.  That is why this is being worked on

shanewilson:
I meant the css rules

bessierogel:
ok. . . my bad. . .I will be quiet.

SidSyrus:
@shanewilson This guy
https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/@ncigdc/uikit/Pagination.js#L34

shanewilson:
that is confusing considering this:
https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/@ncigdc/uikit/Pagination.js#L72

could we try flipping those lines maybe to put active first?

Jephuff:
I don't think it makes sense to apply multiple classes like this 
```
     className={`
       ${styles.tableActionButtons(theme)}
       ${className || (active ? styles.active(theme) : styles.inactive(theme))}
     `}
```
we should probably combine those rules, otherwise we are relying on the order glamor appends styles.

SidSyrus:
@shanewilson Yup, that worked!

PR #1360, opened 2017-03-10
🐛 [PRTL-826] download metadata fix

SidSyrus:
Change the API Url.


cy:
hmm not sure why the GDC_AUTH url wouldn't work, but all metadata shouldn't need auth, it's just using the /files REST endpoint. So if you've tested it and it works then 👍 

aside: https://github.com/NCI-GDC/portal-ui/blob/2704ce661a99cc9c09394a3e192f0aaf3fa46497/modules/node_modules/%40ncigdc/components/MetadataDownloadButton.js#L57 this looks like a typo, wanna fix it while you're here

PR #1365, opened 2017-03-13
🐛 [PRTL-824] Mutation External Reference

SidSyrus:
- Removed Gene External References and added Mutation's.

### DIFF
--- a/modules/node_modules/@ncigdc/components/SSM.js
+++ b/modules/node_modules/@ncigdc/components/SSM.js
@@ -82,14 +82,9 @@ const Mutation = compose(
   const allCasesAggByProject = allCasesAgg.reduce((acc, bucket) =>
     ({ ...acc, [bucket.key]: bucket.doc_count }), {}
   );
-
   const consequences = mutation.consequence.hits.edges;
-  const externalDbIds = consequences.reduce((acc, c) => {
-    if (c.node.transcript.is_canonical) {
-      return c.node.transcript.gene.external_db_ids;
-    }
-    return acc;
-  }, {});
+  const dbSNP = consequences.reduce((acc, c) =>
+    c.node.transcript.annotation.dbsnp_rs ? c.node.transcript.annotation.dbsnp_rs : acc, {});
 
   const functionalImpact = (
     consequences.reduce(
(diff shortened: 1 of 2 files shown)


modules/node_modules/@ncigdc/components/SSM.js
```diff
@@ -210,23 +205,13 @@ const Mutation = compose(
...
-                  <ExternalLink
-                    href={externalReferenceLinks[db](externalDbIds[db][0])}
-                  >
-                    {externalDbIds[db]}
-                  </ExternalLink>
-                ),
-              }))
-          }
+            thToTd={[
+              { th: 'Name', td: 'dbSNP' },
```
Jephuff: I think the ticket meant something like `{ th: 'dbSNP', td: dbSNP }`, so that we can add more in the future, unless there were more instructions outside of the ticket.

modules/node_modules/@ncigdc/components/SSM.js
```diff
@@ -82,14 +82,9 @@ const Mutation = compose(
...
-
   const consequences = mutation.consequence.hits.edges;
-  const externalDbIds = consequences.reduce((acc, c) => {
-    if (c.node.transcript.is_canonical) {
-      return c.node.transcript.gene.external_db_ids;
-    }
-    return acc;
-  }, {});
+  const dbSNP = consequences.reduce((acc, c) =>
+    c.node.transcript.annotation.dbsnp_rs ? c.node.transcript.annotation.dbsnp_rs : acc, {});
```
Jephuff: default should be an empty string instead of an object to match target value.

SidSyrus: Value of `dbSNP` after the reduction would either be a string or `null`

modules/node_modules/@ncigdc/components/SSM.js
SidSyrus: its in the [comment](https://jira.opensciencedatacloud.org/browse/PRTL-824?focusedCommentId=26402&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26402). I will confirm with Bessie.

Jephuff: Yeah definitely confirm, I read that comment as meaning `{ th: 'dbSNP', td: dbSNP }`

bessierogel: Just got a new requirement.  Will update the ticket.
In a nutshell, we will keep the 'External References' table whether of not the rs value exists.
So the condition would be:  If there is a dbSNP_rs value we will display Value = rs#.  If not, we will display Value = 'None'
I will edit ticket

Jephuff: @bessierogel when you say `Name` and `value` do you mean for it to look like this 
image
or would `dbSNP` be on the left and `none` would be on the right?

bessierogel: Requirement had just changed and so, that would have been correct if the requirement did not change.
I will update ticket for the new condition and new UI display.  Or I will post comment here shortly

SidSyrus:
Blocked by [DAT-668](https://jira.opensciencedatacloud.org/browse/DAT-668) due to new UI requirements.

PR #1366, opened 2017-03-13
Prtl 779 removing files from cart

Jephuff:


modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -136,6 +137,39 @@ function fetchFiles (currentFilters, total) {
   };
 }
 
+function fetchFilesAndRemove(currentFilters: ?Object, size: number): Function {
+  return async (dispatch, getState) => {
+    const existingFiles = getState().cart.files;
+
+    if (!existingFiles.length) {
+      dispatch(notify({
+        action: 'full',
```
cy: should probably rename this `action` property, took me a while to figure out where it's used because it's not a redux action, it's just a key into a styles object here https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/uikit/Notification.js#L69 where 'add' has a green background and 'remove' has a red background. So this should be `action: 'remove'`?

cy: oki, added a styles.warning that's yellow

Jephuff: Using the warning now 👍

modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -204,6 +206,39 @@ function fetchFiles(currentFilters, total) {
...
+        component: <Column><span>There are no files in the cart to remove.</span></Column>,
+      }));
+
+      return;
+    }
+
+    const filters = size > 10000 ? replaceFilters({
```
cy: could you use the newly created `MAX_CART_SIZE` const

Jephuff: Yep! done.

PR #1368, opened 2017-03-13
✅🐛 PRTL-778 enforce cart limit in all cart action creators

cy:
- add dux/cart tests

### DIFF
--- a/modules/node_modules/@ncigdc/dux/cart.js
+++ b/modules/node_modules/@ncigdc/dux/cart.js
@@ -10,15 +10,15 @@ import { fetchApi } from '@ncigdc/utils/ajax';
 import { notify } from '@ncigdc/dux/notification';
 import { Column } from '@ncigdc/uikit/Flex';
 import { center } from '@ncigdc/theme/mixins';
-import styled from '@ncigdc/theme/styled';
 
 /*----------------------------------------------------------------------------*/
 
-const UPDATE_CART = 'UPDATE_CART';
-const ADD_TO_CART = 'ADD_TO_CART';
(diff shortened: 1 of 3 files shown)


modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -145,6 +143,13 @@ function addAllFilesInCart(incomingFiles) {
         (existingFile) => existingFile.file_id === file.file_id));
     const filesInCart = incomingFilesArray.length - nextFiles.length;
 
+    if (nextFiles.length + existingFiles.length > MAX_CART_SIZE) {
+      dispatch({
+        type: CART_FULL,
+      });
+      fullNotificationDispatcher(dispatch, nextFiles.length);
```
Jephuff: could `nextFiles.length` be removed? looks like it's not used

modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -52,12 +52,35 @@ const getNotificationComponent = (action, id, notification, dispatch) => ({
...
 });
 
+const fullNotificationDispatcher = (dispatch: Function) => {
+  dispatch(notify({
+    action: 'add',
+    id: `${(new Date()).getTime()}`,
+    component: (
+      <Column>
+        <span>
+        The cart is limited to {MAX_CART_SIZE.toLocaleString()} files.
```
Jephuff: missing an indent :D

modules/node_modules/@ncigdc/dux/cart.js
cy: good catch changed my mind about that function

PR #1370, opened 2017-03-14
🐛 flex div inside btn for firefox

alex-wilmer:


alex-wilmer:
ff doesn't like flex on buttons http://stackoverflow.com/questions/24289251/flexbox-not-working-in-buttons-in-firefox

PR #1372, opened 2017-03-14
🐛 [PRTL-775] Missing "pending toaster" when adding a lot of files to the Cart

SidSyrus:


Jephuff:
Could you add a notification to `fetchFilesAndRemove` as well? it might take awhile too.

SidSyrus:
Sure.

modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -219,6 +225,10 @@ function fetchFilesAndRemove(currentFilters: ?Object, size: number): Function {
       return;
     }
 
+    messageNotificationDispatcher('info',
+      <span>Removing <b>{existingFiles.length}</b> files from cart</span>,
```
Jephuff: This if just removing the files in the query so it should be `size` or better yet, because I have a guard lower to make sure it doesn't try to fetch over `10000` it could say it's removing `Math.min(size, MAX_CART_SIZE)` to match.

SidSyrus: The idea here is to initially show that we are removing whatever number of files we have in cart then go ahead fetch the queried files and only remove those. 
Similar to what we have for adding files in PROD. we will show that we are adding all of the files user currently seeing but then go ahead add only those ones that are not already in the cart. 
What does `size` represent here?

Jephuff: size is the number of files you are removing

SidSyrus: Then I would use `size` over `Math.min(size, MAX_CART_SIZE)`

Jephuff: https://github.com/NCI-GDC/portal-ui/pull/1372/files#diff-e89720563aa6f742cd9e3d9c264e4511R243
Your choice, either have there merits.

SidSyrus: Not showing number of files anymore. Just the message.

PR #1373, opened 2017-03-14
🐛 hookup "back to facet search" button

alex-wilmer:


modules/node_modules/@ncigdc/containers/SmartSearchPage.js
```diff
@@ -94,6 +94,20 @@ class SmartSearchComponent extends React.Component {
           this.Error = null;
 
           this.setQuery();
+
+          document.querySelector('.btn-search-query').onclick = () => {
```
cheapsteak: do we need to worry about removing listener on destroy to avoid leaks (or multiple handlers being fired)?

alex-wilmer: I'll add that, good catch 👍

Jephuff: Is that a problem when using level 0 events? I've never thought of it when using them.

cheapsteak: That's a good point, I'm not sure

cheapsteak: According to this SO http://stackoverflow.com/a/37096563/482053 memory leaks from unattended event listeners aren't a problem anymore ?

Not worth blocking merge of a bug fix in any case, lgtm

PR #1374, opened 2017-03-14
✨ [PRTL-781] show how many files were added to the cart

Jephuff:


modules/node_modules/@ncigdc/dux/cart.js
```diff
@@ -83,10 +83,14 @@ function toggleFilesInCart(incomingFile) {
       return;
     }
 
+    const fileName = incomingFile.file_name || Math.abs(existingFiles.length - nextFiles.length);
```
SidSyrus: I think we should be showing file name if 1 file was already in the cart.

PR #1376, opened 2017-03-14
🐛 don't link from body if no data

alex-wilmer:


modules/node_modules/@ncigdc/components/Home.js
```diff
@@ -145,21 +145,28 @@ const Home = compose(
...
-                    field: 'primary_site',
-                    value: [d._key.replace(/-/g, ' ')],
-                  }],
-                  false
-                )),
-              };
-              push({ pathname: '/projects', query });
+              const key = d._key.replace(/-/g, ' ');
+
+              if (data.find(x => x._key === key)) {
```
Jephuff: We probably shouldn't have a pointer when clicking won't do anything. not sure if that makes sense here or if it would need to be in lolliplot? 
@cy did it here by putting the onclick in the data https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/components/Charts/BarChart.js#L108

alex-wilmer: I think in this case, since they are just inlined raw svgs I can hide any clickable / hoverable items with `display: none` if their related primary site isn't in the data.

Jephuff: that sounds good to :)

PR #1382, opened 2017-03-15
[PRTL-732] show message if no data

SidSyrus:


SidSyrus:
@cy Ready for review.

modules/node_modules/@ncigdc/components/ClinicalCard.js
```diff
@@ -92,99 +92,121 @@ const ClinicalCard = ({
...
-                />
-              }
-              {(!x.treatments || !x.treatments.length) &&
-                <div>No treatments found.</div>
-              }
-            </span>
-          )}
-        />
+                  {(!x.treatments || !x.treatments.length) &&
+                    <div>No Treatments found.</div>
```
cy: Found

modules/node_modules/@ncigdc/components/ClinicalCard.js
```diff
@@ -212,7 +234,11 @@ const ClinicalCard = ({
             />
           }
           {!totalExposures &&
-            <h3>No Exposures found.</h3>
+            <h3
+              style={{ paddingLeft: '2rem' }}
+            >
+              No Exposures found.
```
cy: Found

modules/node_modules/@ncigdc/components/ClinicalCard.js
SidSyrus: I think its because there is already a `<h3>` label above it.
image

PR #1384, opened 2017-03-15
:sparkles: PRTL-845 CurrentFilters expands/hides with ...Less

cy:
screen shot 2017-03-15 at 3 16 18 pm


cheapsteak:
Have a small suggestion for the way expanded logic is handled
How would you feel about something like this?
https://github.com/NCI-GDC/portal-ui/compare/feat/845-current-filters-less-more...feat/845-current-filters-less-more-suggestion?expand=1

cy:
sure, didn't like the `expanded` key everywhere? Merging yours into here

PR #1386, opened 2017-03-15
💄PRTL-810 add maximum width to entity page layout to avoid stretching

cheapsteak:
JIRA:
https://jira.opensciencedatacloud.org/browse/PRTL-810

Test:
Go to an entity page, e.g.
http://localhost:8080/projects/TCGA-CHOL

See that pages are no longer stretched


Jephuff:
It's a bit narrow :D
maybe 1500 or 1600? as the max?

cheapsteak:
updated to 1600px

PR #1393, opened 2017-03-16
🐛 use urlJoin for concating API and GDC_AUTH

Jephuff:


cheapsteak:
As discussed with Jeffrey, 
Love this PR  
Needs a regression test though  
Today isn't a good day to risk breaking things on any page that makes a data request  

Let's keep this open until we have capacity to test

shanewilson:
I'm assuming we don't need this anymore?

Jephuff:
We don't need it, but I think it's still a good idea. 
if we use urlJoin instead of concat then we don't have to worry if the variable ends with a slash or not :)

shanewilson:
okay sounds good.
how come url-join was already in pkg.json?

Jephuff:
this pull request use to be huge, so I started updating when I was already in a file. To avoid unnecessary bugs.

shanewilson:
status?

Jephuff:
rebased, site still works :)

PR #1395, opened 2017-03-17
♿️ PRTL-872 GenePage 508 handle undefineds

cy:


modules/node_modules/@ncigdc/containers/GeneSummary.js
```diff
@@ -24,7 +24,7 @@ const GeneSummaryComponent = ({ node: gene }: TProps = {}) => (
       { th: 'Symbol', td: gene.symbol },
       { th: 'Name', td: gene.name },
       { th: 'Synonyms',
-        td: [].concat(gene.synonyms || []).map(s => <div key={s}>{s}</div>),
+        td: gene.synonyms.length && (gene.synonyms || []).map(s => <div key={s}>{s}</div>),
```
Jephuff: if `gene.synonyms.length` doesn't cause an error, then `gene.synonyms || []` should be needed right?

cy: updated 😛

PR #1397, opened 2017-03-17
🐛 join instead of concat

Jephuff:


cheapsteak:
My bad, should I have let you use url-join

PR #1402, opened 2017-03-20
Fix/846 dl error msg if no projs

cy:
This PR adds an error reducer to show a SessionExpiredModal on 401 or 403 (missing functionality from prod portal), and shows a toast notification if user tries to download token but they have no projects (what the ticket wants).

To test, I did this:
1) Start yarn with
`GDC_AUTH='/api/' yarn start`
because couldn't figure out how to get knit needle to get webpack to also proxy /auth to localhost:5000 :woman_shrugging: 

2) Added
```python
@blueprint.route('/user', methods=['GET'])
def fakeness():
    return jsonify({ "username": "PIKACHU", "projects": {}}), 200


@blueprint.route('/token/refresh', methods=['GET'])
def token_fakeness():
    return jsonify({ 'lol': 'fake token' }), 200
```
to gdcapi/resources/index/v0.py

3) test that projects: {} shows the toast and projects with keys downloads the fake token.

### DIFF
--- a/modules/node_modules/@ncigdc/components/BAMSlicingButton.js
+++ b/modules/node_modules/@ncigdc/components/BAMSlicingButton.js
@@ -50,7 +50,7 @@ const BAMSlicingButton = ({
           file={file}
           closeModal={() => dispatch(setModal(null))}
           setActive={setActive}
-        /> : <NoAccessModal closeModal={() => dispatch(setModal(null))} />))
+        /> : <NoAccessModal />))
     }
   >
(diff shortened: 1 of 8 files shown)


cy:
note: depends on https://github.com/NCI-GDC/authorization/pull/93 else Object.keys(projects) will break when there are no projects. Could check isArray but it's just nicer if we know it's always an Object.

cy:
NCI-GDC/authorization#93 got merged, now this is ready for review for realz

PR #1405, opened 2017-03-20
:lipstick: PRTL-771, PRTL-724 case page icons

cy:
- make ShoppingCart, File, Annotation icons consistent


modules/node_modules/@ncigdc/components/AddToCartButtonAll.js
```diff
@@ -90,14 +90,14 @@ const AddToCartButtonAll = ({ edges, files, total, dispatch }) => (
             aria-label='Add all files to cart'
             style={DropDownItemStyle}
           >
-            <ShoppingCartIcon /> Add all files to the Cart
+            <ShoppingCartIcon style={{ marginRight: '1em' }} /> Add all files to the Cart
```
Jephuff: Should this file use `@ncigdc/theme/icons/ShoppingCart` as well?

cy: yup, forgot - updated!

PR #1411, opened 2017-03-22
:sparkles: MostAffectedCaseTable - # mutations col

cy:
~changing this to one query per row~

### DIFF
--- a/modules/node_modules/@ncigdc/containers/MostAffectedCasesTable.js
+++ b/modules/node_modules/@ncigdc/containers/MostAffectedCasesTable.js
@@ -5,6 +5,7 @@ import Relay from 'react-relay';
 import withSize from 'react-sizeme';
 import { isEqual } from 'lodash';
 import { lifecycle, compose } from 'recompose';
+
 import { parseIntParam, parseFilterParam } from '@ncigdc/utils/uri';
 import Showing from '@ncigdc/components/Pagination/Showing';
 import { withTheme } from '@ncigdc/theme';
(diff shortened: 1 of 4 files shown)


cy:
changed to pre-generated fragments, can't even get a simple string interpolation like $ssmFilter${number} to work inside Relay.QL :/ all fragments written out. Looks dumb, but works. Ready for review @NCI-GDC/oicrdevs

modules/node_modules/@ncigdc/containers/MostAffectedCasesTable.js
```diff
@@ -146,6 +167,183 @@ const MostAffectedCasesTableComponent = compose(
...
+  row18: hits(filters: $ssmsFilters38) {
+    total
+  }
+  row19: hits(filters: $ssmsFilters39) {
+    total
+  }
+}
+`;
+
+const SsmVariables = {
```
Jephuff: definitely not a good idea, but you could generate this list from relay
```
   mutationsFragment40.children.reduce((vars, row) => [...vars, ...row.calls.map(call => call.value.callVariableName)], [])
```
would give you an array of the variable names...
but don't do it haha probably pretty fragile.

alex-wilmer: hmm couldn't this be

```
_.range(0, 40).reduce((p, c) => ({ ...p, [`ssmsFilters${c}`]: null ]), {})
```

cy:
ok, ok `key={caseId}` on the `<MutationsCount>`s made the page change work. Thanks for the help @alex-wilmer. I still don't like it but I'll live.

PR #1415, opened 2017-03-23
✨ [PRTL-893] arrange columns

alex-wilmer:
WIP


shanewilson:
invalid?

alex-wilmer:
yeah this isn't the path forward, but definitely still a guide for how to start doing it once we come back top it.

PR #1420, opened 2017-03-23
🐛 try to lock down yarn version by overwriting travis version

Jephuff:
~this might fix yarn failing~
~it did not~
now it did


cy:
😭

PR #1421, opened 2017-03-24
🐛 [PRTL-824] Mutation external reference

SidSyrus:
reading `dbSNP` and `COSMIC` ids as external reference for mutation.


modules/node_modules/@ncigdc/components/SSM.js
```diff
@@ -210,25 +206,26 @@ const Mutation = compose(
             alignSelf: 'flex-start',
           }}
         />
-        {externalDbIds && !!Object.keys(externalDbIds).length &&
```
cy: you're sure that the other external reference links, Entrez Gene etc are supposed to be deleted?

SidSyrus: Ya those were gene references in mutation entity page

cy: ok looks good then

PR #1422, opened 2017-03-24
🔧 generate schema on start, remove from vc

alex-wilmer:
The intention here is that anytime the API graphql schema is updated, we need to update the UI anyways. This PR makes sure a fresh relay schema is created whenever the portal is booted up.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -11902,22 +11902,6 @@
             {
               "args": [],
               "deprecationReason": null,
-              "description": "When paginating forwards, are there more items?",
-              "isDeprecated": false,
-              "name": "hasNextPage",
-              "type": {
-                "kind": "NON_NULL",
-                "name": null,
-                "ofType": {
-                  "kind": "SCALAR",
-                  "name": "Boolean",
-                  "ofType": null
-                }
-              }
-            },
-            {
-              "args": [],
-              "deprecationReason": null,
               "description": "When paginating backwards, are there more items?",
               "isDeprecated": false,
               "name": "hasPreviousPage",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -69,7 +69,7 @@
   "scripts": {
     "knit": "knit",
     "build": "knit build --skip-preflight --verbose",
-    "start": "knit server",
+    "start": "rm data/schema.json&&rm data/version.json&&knit schema&&knit server",
```
Jephuff: This will fail if the files don't already exist because rm will fail. probably should replace `&&` with `;`

package.json
```diff
@@ -69,7 +69,7 @@
   "scripts": {
     "knit": "knit",
     "build": "knit build --skip-preflight --verbose",
```
Jephuff: Do we need to do it in build as well?

package.json
Jephuff: A thought, `knit schema` should overwrite the existing files right? 
a command like `knit schema && knit server` might be better because then it will fail to start if you can't get the schema, but it doesn't delete the schema so you can still edit the command and run with the existing files.

alex-wilmer:
note: keep in vc

alex-wilmer:
@shanewilson @Jephuff is this still helpful? tried to change my api schema and things will work, but I'm not sure I totally understand the caching at this problem. at least this generates a schema every time

package.json
shanewilson: I don't think so - build might be used by a CI or tungsten or something.

data/schema.json
```diff
@@ -12312,18 +12328,6 @@
               "deprecationReason": null,
               "description": null,
               "isDeprecated": false,
-              "name": "ssm_tested",
```
shanewilson: this went away?

alex-wilmer: yea it changed to `available_` something something

Jephuff: yeah, it changed to `available_variation_data: ['ssm']`

package.json
Jephuff: Wouldn't we want this for tungsten to ensure we always have the correct schema? Otherwise we run the risk of having changes that are only on collab?

PR #1423, opened 2017-03-24
🐛 PRTL-859 make custom facets w/ data work again by increasing timeou…

cheapsteak:
…t and guarding against blank aggregations


cy:
sure, it'll still break in IE because IE has a default time out of 60s and user needs to change the registry to change it (https://support.microsoft.com/en-us/help/813827/how-to-change-the-default-keep-alive-time-out-value-in-internet-explorer) orz - but I believe this was an issue we never fixed in prod, had two attempts at creating a custom endpoint https://github.com/NCI-GDC/gdcapi/pull/490 and https://github.com/NCI-GDC/gdcapi/pull/568 The speed ups weren't considerable & we never merged them

PR #1425, opened 2017-03-24
🐛 Quicksearch handle when field value is array

cheapsteak:


modules/node_modules/@ncigdc/components/QuickSearch/QuickSearchResults.js
```diff
@@ -115,7 +115,7 @@ const doctypeFieldsOfInterestMap = {
 
 const ResultHighlights = ({ item, query, style }) => {
   const fields = doctypeFieldsOfInterestMap[item._type] || [];
-  const values = fields.map(field => (_.isFunction(field) ? field(item, query) : _.get(item, field)));
+  const values = fields.map(field => (_.isFunction(field) ? field(item, query) : [].concat(_.get(item, field)).join(',')));
```
Jephuff: maybe give the get a default value of `[]` so we don't end up showing the user `undefined`

PR #1427, opened 2017-03-27
PRTL-902 Set height of cancer distribution histogram to 300. Remove bar width scaling

cheapsteak:
Have confirmed that we indeed want to remove bar width scaling based on number of bars 
https://oicr-gdc.slack.com/archives/C0H0726P7/p1490623679788124

### DIFF
--- a/modules/node_modules/@ncigdc/containers/FrequentMutationsChart.js
+++ b/modules/node_modules/@ncigdc/containers/FrequentMutationsChart.js
@@ -41,9 +41,10 @@ const FrequentMutationsChartComponent = compose(
 }) => {
   const chartMargin = { top: 30, right: 50, bottom: 105, left: 40 };
 
-  const bandWidth = ssms && (
+  const maxBars = 20;
+  const bandWidth = (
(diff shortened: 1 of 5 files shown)


alex-wilmer:
maybe this is a good opportunity to remove that code from the individual charts? would it be possible to handle this stuff one time in `charts/BarChart.js` ?

cheapsteak:
I'd recommend keeping it here for a release or until we're certain that this spec wont get reverted

modules/node_modules/@ncigdc/containers/FrequentMutationsChart.js
```diff
@@ -41,9 +41,10 @@ const FrequentMutationsChartComponent = compose(
 }) => {
   const chartMargin = { top: 30, right: 50, bottom: 105, left: 40 };
 
+  const maxBars = 20;
   const bandWidth = ssms && (
```
alex-wilmer: can remove these null checks

cheapsteak: removed

PR #1428, opened 2017-03-27
✨ [PRTL-907] hide viz when ssm_tested isn't true

Jephuff:
depends on https://github.com/NCI-GDC/gdcapi/pull/741


shanewilson:
isn't ssm tested the thing that is turning into an array? when is that happening?

Jephuff:
Yep it is, looks like the index was updated, I will update the PR to use that field.

PR #1437, opened 2017-03-27
✨♻️💄PRTL-901 Issues on the Cart - reproduce same behaviour as in Prod 🙇🏻🙇🏻🙇🏻🛒

cheapsteak:
Addresses following tickets related to cart
https://jira.opensciencedatacloud.org/browse/PRTL-858
https://jira.opensciencedatacloud.org/browse/PRTL-886
https://jira.opensciencedatacloud.org/browse/PRTL-906
https://jira.opensciencedatacloud.org/browse/PRTL-884
https://jira.opensciencedatacloud.org/browse/PRTL-885

look for "fix-901-cart" on deploybot


modules/node_modules/@ncigdc/utils/formatFileSize.js
```diff
@@ -1,31 +1,4 @@
 // @flow
+import filesize from 'filesize';
```
cy: why not delete this thin wrapper and call `filesize()` directly instead of `formatFileSize`. Or maybe use `/components/FileSize.js`? Could you filesize this as well https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/components/File.js#L143 while you're at it

cy:
the text becomes weirdly selected in `SummaryCard` when switching back from pie to table
screen shot 2017-03-28 at 5 36 50 pm
it's also like that in `relay` maybe no one cares :woman_shrugging:

shanewilson:
we are getting rid of pie charts on those tables. did that not happen yet? it was mentioned before I left, maybe got lost.

modules/node_modules/@ncigdc/utils/formatFileSize.js
cheapsteak: It happened to serve as an abstraction layer similar to the `<XXLinks>`s and our `withRouter`

I'm uncertain about how concrete the specs would remain for filesize formatting. We used to use base 10 (explicitly opting for it in an instance) to calculate filesize but showed `KB` (signaling base 2) instead of `kB` (signaling base 10) as the unit; there was one place that was directly using filesize instead of this utility function, and thus showed the correct unit format of `kB` instead of `KB`. The ticket asked that the `kB` be changed to `KB`, potentially unaware that those are 2 different units  

I asked in slack but no one responded so I made an out-of-spec assumption that it's not a big deal to change from using base 10 to base 2 to satisfy the ticket's requirement of having uppercase units, which also allows our desired result to happily align with the default options of the library; but if specs change (either it's important to use base 10, or perhaps even important to use base 10 and have uppercase units), it'd be convenient to have a single place where modifications need to be made

It's not a DAL I would have created starting out from scratch, but the cost/benefit seemed to work out in favour of keeping the module and avoiding bloating the PR's diff. 

Don't feel strongly about it though, don't mind replacing all instances if we're fairly certain this format will remain acceptable :)

cheapsteak:
Have not seen any tickets regarding removing pie charts

modules/node_modules/@ncigdc/utils/formatFileSize.js
shanewilson: Can you confirm the file sizes we show on the site are the same as when you download the file?

cheapsteak:
The selection bug is indeed annoying, but it only seems to happen if you click quickly enough for the browser to register it as a double click

Perhaps we could force a blur when flipping between these card states but perhaps not worth looking into if we're getting rid of the pies

modules/node_modules/@ncigdc/utils/formatFileSize.js
cheapsteak: We show `51 KB`


osx shows

image

shanewilson: did you check production or this branch?

cheapsteak:
uat

modules/node_modules/@ncigdc/utils/formatFileSize.js
cheapsteak: I checked on UAT  
I don't think it'd break anything, but if we do get feedback saying "this used to be 54 and now it's 51" this module will be here to make the switchover easy

shanewilson: Do we use `formatFileSize.js` anywhere? If not we should remove it but if we do `components/FileSize.js` should use it instead of `filesize` directly.

cheapsteak: Not sure how to test?   

Some places use base 2 (linux, windows, osx before snow leopard apparently), some use base 10 (harddrive vendors and newer osx maybe?) If lidos feels strongly about having it displayed a certain way then let's go with whatever that is (iirc base 2 is more typical). That information was absent in the ticket so I went with the library default  

There isn't really a wrong way to display it, base 10 with `kB` and base 2 with`KB` are more "technically correct", base 10 with `KB` is fine too (what we had before), I doubt scientists care about the nuances of kilo vs kibi

We just need to find out what format they want it in and set it to that. If we do end up wanting base 10 with KB though and we have the module, we can just change `export default filesize` to `export default (input) => filesize(input, {base: 10}).toUpperCase()`

shanewilson: This isn't a theoretically conversation - this site is in production already. They way it worked was the way we wanted it.

cheapsteak: Was `formatFileSize.js` previously conforming to spec? If that's the case then we could either do `filesize(bytes, {base: 10}).toUpperCase()` and export that here, or I could revert this file and have places that are using the filesize module use this instead (and remove the dependency on filesize)

cheapsteak: Hadn't intended to change anything, my agenda's usually to change as little as possible, but saw some places use formatFileSize and another place use filesize, the place using filesize was newer than the place using formatFileSize, figured if we're going to go with the library then we can get rid of our self-rolled formatter, and maybe the somewhat nonstandard units were by accident and went with the one from the library

cheapsteak:
Set formatting to use base 10 with uppercase units :)
Are we good to merge?

modules/node_modules/@ncigdc/routes/CartRoute.js
```diff
@@ -34,6 +35,15 @@ export default connect(state => (state.cart))(
       Container={CartPage}
       queryConfig={new CartRoute(props)}
       environment={Relay.Store}
+      render={({ error, props }) => {
```
cy: is it possible to have `<Loader />` be in the CartPage `<FileTable />` instead of the route so that the SummaryCards and CountCards don't need to reload on page change?

cheapsteak: Unfortunately not, nothing inside the container seems to be rendered until relay gets its data

cy: what about moving the `FileTable` into its own `CartFileTable` container, kinda like the viz tables ie `containers/FrequentMutationsTable`

cheapsteak: My hunch is the readystate would bubble up through containers but I'm not sure
@alex-wilmer do you know?

alex-wilmer: why did we change to using Loader here instead of the red bar?

alex-wilmer: and to answer your question.. as long as the filetable triggers its own query via `setVariables` and `@include` then it can be in its own Loader and the summary cards query can return first.

cheapsteak: The red bar did not exist at the time of this PR, I think now it has both
I'll get rid of the loader then

cheapsteak: Spinner loader has been removed

cy: The red bar acts strangely a lot of the time though, like it'll load forward and then bounce back before reaching the end of the page. Happens a lot, like during a demo once and no one liked it. If the FileTable could get it's own query & spinner that'll be ideal. Hey it's not as bad as each Mutation Count row being it's own container.

alex-wilmer: haha it's weird but it makes 100% sense if you look at the component. it just takes a value from 0 - 100 to figure out its width. if someone wants to play with it to make it go to the other side or listen to an onfinished event to do something, go for it. people forget that before the red bar there was nothing and it looked like the site was freezing constantly

cheapsteak: Discussion seems to be gradually going off scope of items in this ticket :)  

I recommend that we create a ticket for fixing behaviour of red loader, and a ticket to create loading state indicators for the FileTable. iirc Alex and Shane wanted a skeleton state, maybe something like ICGC's projects table

image

cy: fair enough

PR #1441, opened 2017-03-28
[PRTL-923 & PRTL-924] Entity page icons

SidSyrus:
Using the initial of the entity type as icons to appear on the upper left hand side of an entity page


modules/node_modules/@ncigdc/components/QuickSearch/QuickSearchResults.js
```diff
@@ -149,7 +147,9 @@ export default ({ results = [], query, isLoading, onSelectItem, onActivateItem }
           onClick={() => onActivateItem(item)}
         >
           <div style={styles.itemIconWrapper}>
-            <ResultIcon type={item._type} style={styles.itemIcon} />
+            <span style={styles.itemIcon}>
+              {getIconClassNameByEntityType(item._type)}
```
cheapsteak: `getIconClassNameByEntityType` this function should probably be renamed as it's no longer returning a class name

cheapsteak:
`getIconClassNameByEntityType` is also used in `FacetSelection.js`

PR #1442, opened 2017-03-28
🐛 protect against page not-loading on occasions when item.file_count is not a number

cheapsteak:


alex-wilmer:
I'm not opposed to this, but it kinda makes it look like it can be something other than a number and something other than null / undefined. I think existence check here would be good enough?

Jephuff:
might be better to do something like `(item.case_count || 0).toLocaleString()` or `parseInt(item.case_count || 0, 10).toLocaleString()`. that way no empty link, which means no 508 :D

cheapsteak:
switched to `(item.case_count || 0)`

PR #1446, opened 2017-03-28
✨ use Lolliplot relay container in SSMPage

alex-wilmer:


alex-wilmer:
I'm getting a mismatched variables error here.. I think it's because the viewer query will run regardless of the `fetchGene` variable? not sure

modules/node_modules/@ncigdc/components/SSM.js
```diff
@@ -289,19 +271,16 @@ const Mutation = compose(
...
-            transcripts={gene.transcripts.hits.edges.map(x => x.node)}
-            selectTranscript={setActiveTranscript}
-            data={lolliplotData}
-          />
-        </Column>
-      )}
+      <Column style={{ ...styles.card, marginTop: '2rem' }}>
+        <Lolliplot
+          mutationId={mutation.ssm_id}
+          geneId={canonicalGeneId}
```
Jephuff: So this is the part causing the relay error. Looks like we can't have props that are the same name as relay variables.

alex-wilmer: oh doh >< yeah I knew that but didn't notice it here. thanks!

Jephuff:
@alex-wilmer I linted some as well... want to take  a look?

PR #1447, opened 2017-03-29
Eslint rules update

Jephuff:
proposed change to make some rules warnings instead of errors to make it a little clearer.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,15 +2,7 @@
   "extends": ["@knit/socks-react"],
   "rules": {
     "flowtype/no-weak-types": 0,
-    "object-curly-spacing": 1,
-    "comma-dangle": 1,
-    "keyword-spacing": 1,
-    "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
     "fp/no-this": 0,
-    "no-invalid-this": 2,
-    "comma-spacing": 1,
-    "max-len": 1,
-    "import/extensions": 1,
-    "import/first": 1
+    "no-invalid-this": 2
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,16 @@
 {
   "extends": ["@knit/socks-react"],
   "rules": {
-    "flowtype/no-weak-types": 0
+    "flowtype/no-weak-types": 0,
+    "object-curly-spacing": 1,
+    "comma-dangle": 1,
+    "keyword-spacing": 1,
+    "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
+    "fp/no-this": 0,
+    "no-invalid-this": 2,
+    "comma-spacing": 1,
+    "max-len": 1,
+    "import/extensions": 1,
+    "import/first": 1
   }
 }


alex-wilmer:
I would still like almost most of these things to be errors

shanewilson:
no warnings - either remove the rule or keep it as an error - making things warnings just means we will have to deal with warning spam.

shanewilson:
also where did this list come from? can you provide reasons why you want to disable each rule?

alex-wilmer:
Specifically, for me, I would like to keep the following:

```
"rules": {
    ...
    "object-curly-spacing": 1,
    "comma-dangle": 1,
    "keyword-spacing": 1,
    "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
    "comma-spacing": 1,
    "max-len": 1,
    ...
}
```

Jephuff:
I just in general think if it's not going to cause errors, then it shouldn't be an error. like `object-curly-spacing` is just a style preference so well I'm coding, I really don't need to fix that right away, but if I use an undefined variable I need to know well I'm coding so I can fix it before opening the browser. 

this list of rules are all current errors in our project that I've run into. 

I also added this
```
"fp/no-this": 0,
"no-invalid-this": 2,
```
because we use `this` all over the place 

But If I'm the only one then I will close this PR :D

alex-wilmer:
I'm certainly in favor of removing to `this` related rules, until such a time that we're never needing it (like once we've completely migrated over to reason)

shanewilson:
people ignore warnings - which basically makes them useless and spams linter output.
you can setup your editor to auto eslint-fix so those style warnings shouldn't be a problem.

shanewilson:
I'm fine with disabling `fp` rules though

shanewilson:
you can't ignore errors cause the CI won't pass. eslint doesn't fail on warnings though.

shanewilson:
> `"no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }]`

what behaviour are you trying to disable here? or are you adding a check?

Jephuff:
It allows unused vars when using rest
so with this change, this wouldn't be an error
```
({a, ...props}) => props
```

cheapsteak:
Making style offenses red squiggly errors significantly dilutes the usefulness of errors for debugging

I've found that I've started to ignore red squiggles (when reading existing code) because most of the time it's giving me useless information; and when I'm trying to debug something, I've been annoyed that when I hover over the red squigglies that are supposed to be helpful in telling me where the errors might be coming from are just saying the line is too long

Using auto fixers in the editor is not a great solution because single character changes in files results in PRs with diffs in places that changed only stylistically

Would be okay with using something like https://github.com/prettier/prettier

cheapsteak:
I don't like the pattern of deconstructing an argument solely to remove it from being spread   
It's too inexplicit, taking advantage of a side effect  
In functions with a few destructured arguments, how can you tell which ones are there because it needs to be omitted from something that's using it inside the function, and which ones were just forgotten to be removed?

shanewilson:
`prettier` still just auto formats code so that doesn't really fix:

> Using auto fixers in the editor is not a great solution because single character changes in files results in PRs with diffs in places that changed only stylistically

Most of the files have a billion red squiggly lines because they come from the ng version and nobody  fixed the errors. The code that started in the relay version are clean and the errors are very useful.

Also auto-fixing with eslint or prettier means you shouldn't have to worry about style changes, just save the file and they mostly fix themselves.

cheapsteak:
Prettier will allow us to have one single PR that fixes everything throughout the repo, eslint autofixing via commandline would probably be able to do the same? (haven't used that before) eslint via editor plugin can only fix one file at a time, tainting the diff containing meaningful changes with meaningless style fixes

shanewilson:
that seems unrelated to this discussion though - since this is about which eslint rules we want not how to fix files in bulk.

alex-wilmer:
There's missing context here... PRs that don't follow the rest of the style don't pass Travis normally but since we've been on relay branch here for so long we've been able to get away with it.

Personally I would not like to see a lot of `// eslint-disable-line` just to make travis pass once we get to that stage.. that's why I'm in favor of removing some of the fp rules there's no way to get around.. but I don't want this codebase to eventually contain many different coding styles, including my own weird ones.

`eslint --fix` works great for me, except in some scenarios where it's shifting giant blocks of JSX and i'm not sure why.

cheapsteak:
This is definitely related. I'd say it's not about which rules we want but whether they should be green or red squigglies

We're currently handicapping our debugging abilities, depriving ourselves of the ability to differentiate between squigglies that cause errors and stylistic squigglies. Currently red squigglies are kind of like the boy who cried wolf

An initial single pass and also continuously auto-enforcing prettier or eslint autofix would solve the problem of code style issues a different way and make red squigglies meaningful again

alex-wilmer:
hmm can we make tests fail on warnings as well as errors?

cheapsteak:
That is useful context

I think if this is being done automatically, the trade off of oddly indented jsx is worth having useful errors and not worrying about missing commas breaking build

alex-wilmer:
I think we should fiddle with the jsx rules a bit before autofixing the entire codebase otherwise we're bound to "unfix" them by accident as we code.

cheapsteak:
I'm recommending that no code make it into version control without being processed by the style fixer

Perhaps by adding this as a pre-commit hook

shanewilson:
that is what the eslint rules are for and why I don't want to remove them ... :persevere:

cheapsteak:
which comment are you referring to?

shanewilson:
The fact that many of the files have many errors is itself a bug - it is not standard operating procedure so I'm not going to change our workflow or rules to accommodate that.

Jephuff:
If the only issue with warnings is it doesn't fail in travis, why not make it fail on warnings in travis? Then jerks like me who hate auto formatting can still get the benefit of green squiggles :D

shanewilson:
travis fails when eslint fails

Jephuff:
yeah, so if we add `--max-warnings 0` to eslint when we run it, it will error if there are warnings

PR #1456, opened 2017-03-30
🐛 PRTL-942 update logic for choosing transcript

cheapsteak:
Before:
image

After:
image


shanewilson:
what is MO?

shanewilson:
oh modifier

PR #1458, opened 2017-03-31
⚡️ PRTL-744 project page performance tweaks

cheapsteak:
Higher green lines (FPS) is better, shorter yellow blocks is better, shorter periods of low FPS is better, fewer solid red rectangles at the top is better 

Before:
image

After:
image

Stats (note: using timeline slows absolute performance)
Before: ~10.7s of 0 FPS, 13.8s total scripting time
After: ~4.7s of 0 FPS, 7.3s total scripting time

Should result in a shorter period where user interaction feels jolt-y

### DIFF
--- a/modules/node_modules/@ncigdc/components/ProjectVisualizations.js
+++ b/modules/node_modules/@ncigdc/components/ProjectVisualizations.js
@@ -118,7 +118,7 @@ const ProjectVisualizations = enhance(({
         </h1>
         <Column>
           <Row>
-            <Column flex="1">
+            <Column flex="none" style={{ width: '50%' }}>
               <FrequentlyMutatedGenesChart
                 defaultFilters={fmFilters}
                 projectId={projectId}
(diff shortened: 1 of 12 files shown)


modules/node_modules/@ncigdc/containers/MostAffectedCasesTable.js
```diff
@@ -45,7 +45,7 @@ const MostAffectedCasesTableComponent = compose(
     },
   }),
   withTheme,
-  withSize()
+  withSize({ refreshRate: 200 })
```
shanewilson: could we give `withSize` sane defaults so we don't have to specify each time?

modules/node_modules/@ncigdc/components/ProjectVisualizations.js
```diff
@@ -118,15 +118,15 @@ const ProjectVisualizations = enhance(({
         </h1>
         <Column>
           <Row>
-            <Column flex="1">
+            <Column flex="none" style={{ width: '50%' }}>
```
shanewilson: why not flex?

modules/node_modules/@ncigdc/containers/MostAffectedCasesTable.js
cheapsteak: Make our own module and import from there?

cheapsteak:
Made own module with default refresh rate

cheapsteak:
Good to merge? @shanewilson

PR #1462, opened 2017-03-31
:bug: PRTL-960 cosmic cosm/cosn link change

cy:


modules/node_modules/@ncigdc/components/SSM.js
```diff
@@ -234,7 +234,7 @@ const Mutation = compose(
                 (c) => (
                   <li key={c}>
                     <ExternalLink
-                      href={externalReferenceLinks.cosmic(c.match(/(\d+)$/g))}
+                      href={externalReferenceLinks[c.substring(0, 4).toLowerCase()](c.match(/(\d+)$/g))}
```
shanewilson: what is the value of `c`?

cy: either `COSN{\d+}` or `COSM{\d+}`

PR #1464, opened 2017-03-31
:bug: PRTL-892 link source files to repo page

cy:
~~wip~~ done


cy:
Ready for review. Includes all child tickets of this https://jira.opensciencedatacloud.org/browse/PRTL-932

Does deeplinking as in this module https://github.com/rafrex/react-router-hash-link/blob/master/src/index.js using MutationObserver, but that module did not account for our header height, so pulled in @Jephuff's https://github.com/NCI-GDC/portal-ui/pull/1242, massaged in MutationObserver and worked into InternalLink. Seems to work, and doesn't seem to break anything else.

Jephuff:
Looks good, I think there is an extra commit from Alex here, maybe rebase?

cy:
@Jephuff - how did that get there? rebased.

PR #1470, opened 2017-04-03
✨PRTL-980 add ssm chart & table in mutations tab

cheapsteak:


alex-wilmer:
lgtm

modules/node_modules/@ncigdc/utils/survivalplot.js
```diff
@@ -20,8 +20,8 @@ export const enoughData = (data: Object) => (
   data.results.length && data.results.every(r => r.donors.length >= MINIMUM_CASES)
 );
 
-async function fetchCurves(filters: ?Array<Object>): Promise<Object> {
```
shanewilson: where does 25 come from? Is that all we normally ask for?

cheapsteak: 30 makes it time out some of the time

cheapsteak: I'll add a comment

shanewilson: the default in the api is 1000 though right? is 25 even enough to make a useful chart?
we probably need to do something about that speed

cheapsteak: Useful as a poc/mvp-demo, but I imagine a researcher would expect the graph to either show all results with 0 truncation, or use some kind of sampling that would result in a graph that is representative of the population

It's a problem we should definitely solve before putting this into production

PR #1472, opened 2017-04-03
✨ use newer case aggs in cohort

alex-wilmer:


cheapsteak:
@alex-wilmer not sure why github is reporting such a large diff, most additions seem to already be on the relay branch, would a rebase help?

alex-wilmer:
rebased, should be good now? not sure why the schema file keeps changing on me, but we can discard that if its' wrong

shanewilson:
UI will change anytime the api changes even no schema changes are done because we save the api schema version. Should probably make it so we don't over write if the schema doesn't change?

alex-wilmer:
I think we might need to remove some stuff? the "add custom filters" stuff won't work I don't think

cheapsteak:
Seems to work?
http://deploybot.gdc.cheapsteak.net:33359/cohort

Although maybe we should remove it anyway since it'll be the only tab that has it (genes and mutations do not)

alex-wilmer:
Well... I guess this is further along than what was there before. Probably a better jumping off point if we need to change anything. I guess I'll merge this for now since we still have some lead up time to figure things out

PR #1477, opened 2017-04-05
🐛 [PRTL-995] update schema for mutation page

Jephuff:


shanewilson:
this is done in relay

PR #1479, opened 2017-04-05
Various table / chart fixes

alex-wilmer:


alex-wilmer:
number of corrections + using `score` instead of filtered cases on the node since this is the same value.

also this updates `withFilters` to take some optional args and return `null` for filters by default

PR #1482, opened 2017-04-05
React lolliplot

alex-wilmer:
wip


alex-wilmer:
okay! this is ready for review now. the main thing to check here is that when using the quicksearch while on gene or mutation pages, the chart updates accordingly

PR #1483, opened 2017-04-06
🐛 guard against null gene_strand

alex-wilmer:


alex-wilmer:
saw this here: https://gdc-portal-staging.datacommons.io/ssms/92b75ae1-8d4d-52c2-8658-9c981eef0e57

PR #1486, opened 2017-04-06
Fix/994 gene pg cancer dist

cy:
~wip~

### DIFF
--- a/modules/node_modules/@ncigdc/containers/CancerDistributionTable.js
+++ b/modules/node_modules/@ncigdc/containers/CancerDistributionTable.js
@@ -1,79 +1,130 @@
 // @flow
 import React from 'react';
 import Relay from 'react-relay';
-import { buildCancerDistributionData } from '@ncigdc/utils/data';
+import { lifecycle, compose } from 'recompose';
+import { groupBy, isEqual, head, sortBy } from 'lodash';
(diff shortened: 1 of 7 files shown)


modules/node_modules/@ncigdc/containers/CancerDistributionTable.js
```diff
@@ -1,32 +1,72 @@
...
 type TProps = {|
   node: Object,
   aggregations: Object,
 |};
 
-const CancerDistributionTableComponent = ({
+const CancerDistributionTableComponent = compose(
+  lifecycle({
+    componentDidMount(): void {
+      this.props.relay.setVariables({
```
shanewilson: do you need this? relay query can't see props?

alex-wilmer: I don't think so http://stackoverflow.com/questions/34255662/props-in-relay-preparevariables

shanewilson: these ajax containers would probably be easier if they were renderers or at least routes so they can pull in the route params

shanewilson:
how is this coming?

cy:
need to make the same changes in the CancerDistributionChart

cy:
@shanewilson this is ready for review

PR #1493, opened 2017-04-07
Fix/ssm cancer dist

cy:


cy:
@shanewilson ready for review, also adds `{ field: 'available_variation_data', value: 'ssm' }` to filter

PR #1495, opened 2017-04-07
:sparkles: PRTL-1007 add variant_caller to Mutation facet panel

cy:


modules/node_modules/@ncigdc/containers/cohort/SSMAggregations.js
```diff
@@ -22,6 +22,7 @@ const presetFacets = [
   { title: 'Consequence Type', field: 'consequence.transcript.consequence_type', full: 'ssms.consequence.transcript.consequence_type', doc_type: 'ssms', type: 'terms' },
   { title: 'Mutation Type', field: 'mutation_type', full: 'ssms.mutation_type', doc_type: 'ssms', type: 'terms' },
   { title: 'Data Categories', field: 'occurrence.case.summary.data_categories.data_category', full: 'ssms.occurrence.case.summary.data_categories.data_category', doc_type: 'ssms', type: 'terms' },
+  { title: 'SSM Callers', field: 'occurrence.case.observation.variant_calling.variant_caller', full: 'ssms.occurrence.observation.variant_calling.variant_caller', doc_type: 'ssms', type: 'terms' },
```
Jephuff: I don't think this field exists. Maybe it should be 
`ssms.occurrence.case.observation.variant_calling.variant_caller`

cy: yup, fixed. thanks!

PR #1508, opened 2017-04-10
✨ PRTL-1024 PRTL-1031 Add quicksearch to home page 

cheapsteak:
Uses namespacing solution found by Jeff and discussed here https://github.com/acdlite/recompose/issues/358

### DIFF
--- a/modules/node_modules/@ncigdc/components/Home.js
+++ b/modules/node_modules/@ncigdc/components/Home.js
@@ -17,6 +17,7 @@ import { makeFilter } from '@ncigdc/utils/filters';
 import styled from '@ncigdc/theme/styled';
 
 import ExploringLinks from './ExploringLinks';
+import HomeSearch from './HomeSearch';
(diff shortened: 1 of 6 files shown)


cheapsteak:
@alex-wilmer could you take a look? I think you'll like it :)

alex-wilmer:
Interesting. was the namespacing required to make something work? looks like you're just destructuring the stuff you had just previously namespaced anyways. it's a good idea to avoid collisions though definitely. 

lgtm besides missing type definition `TSearchHit`

cheapsteak:
Preventative measure, especially for props like `state`, `setState`, and `reset`

cheapsteak:
@alex-wilmer TSearchHit type imported :)

alex-wilmer:
Did you push? I don't see any changes

cheapsteak:
my bad, pushed to the wrong branch, should be there now

PR #1509, opened 2017-04-10
Fix/1012 add in filters bug

cy:


shanewilson:
why do we need addInFilters? It looks like toggleFilters but it allows duplicates?

cy:
Yes it's the same as `toggleFilters` but it does not toggle, as in it doesn't remove an existing value. https://github.com/NCI-GDC/portal-ui/pull/1509/commits/51d93118e65d50b509bd49bcaf35e915e2291b43#diff-ab3647c4fa0d85b323f79a06c47f016cR94 tests to show the difference. 

screen shot 2017-04-11 at 10 03 14 am
It's used in the dropdown of the autocomplete. Copies behaviour of prod, also think it makes more sense then toggling here.

PR #1511, opened 2017-04-11
🎛 Increase default number of cases to fetch for survival plot to 500

cheapsteak:
Doesn't seem to time out anymore
I wonder what was causing this to time out before
Worst case scenario (0 filters) should also be cached by varnish so this may work

### DIFF
--- a/modules/node_modules/@ncigdc/utils/survivalplot.js
+++ b/modules/node_modules/@ncigdc/utils/survivalplot.js
@@ -20,8 +20,7 @@ export const enoughData = (data: Object) => (
   data.results.length && data.results.every(r => r.donors.length >= MINIMUM_CASES)
 );
 
-// default size is 25 because 30 sometimes makes the API's ES query time out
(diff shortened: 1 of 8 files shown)


modules/node_modules/@ncigdc/utils/survivalplot.js
```diff
@@ -21,7 +21,7 @@ export const enoughData = (data: Object) => (
 );
 
 // default size is 25 because 30 sometimes makes the API's ES query time out
```
cy: out of date comment 😱

cheapsteak: nice catch, fixed and pushed

cy:
Didn't rely on the default in the click handler and passed it in explicitly https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/containers/FrequentMutationsTable.js#L327 & https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/containers/FrequentlyMutatedGenesTable.js#L216

PR #1513, opened 2017-04-11
🐛 [PRTL-1028] fix survival case count

Jephuff:


modules/node_modules/@ncigdc/components/ProjectVisualizations.js
```diff
@@ -49,24 +48,25 @@ const enhance = compose(
...
-    if (!numCasesAggByProject) {
-      return;
+  withPropsOnChange(
+    ['projectId', 'caseCount', 'survivalPlotSampleSize'],
+    async ({ projectId, setState, caseCount, survivalPlotSampleSize }) => {
+      if (!projectId) return;
+
+      const defaultSurvivalData = await getDefaultCurve({
+        totalCases: caseCount,
+        currentFilters: { op: '=', content: { field: 'cases.project.project_id', value: projectId } },
```
cy: should this also take in the clinical filters like https://github.com/NCI-GDC/portal-ui/blob/relay/modules/node_modules/%40ncigdc/routes/ProjectRoute.js#L29 cuz the ticket says this 
> I believe the Survival plot title should show 974 cases instead of 986 = number of cases for which we have clinical data needed for the analysis (as opposed to number of donors with ssm data)

Jephuff: Oop, my bad. I missed that part 😳

Jephuff: Ok I think it's better now :) 
I just use the number returned from the endpoint to be consistent with the number thats used when splitting into 2 lines

PR #1525, opened 2017-04-11
name as survival analysis

alex-wilmer:


cheapsteak:
Is this just for the demo?  
I'm not sure about the UX on this  
The behaviour seems rather unintuitive, people would not expect clicking the name would be related to the graph (name is an identifier, would expect the same behaviour as clicking on ID)

alex-wilmer:
> Is this just for the demo?

yes

shanewilson:
who asked for this change?

shanewilson:
oh its an attempt to prevent scrolling - ok

PR #1526, opened 2017-04-12
💄 column widths, toLocalString, survival button style

alex-wilmer:
Figured it was better to make something we'll end up keeping

image


alex-wilmer:
@cheapsteak @shanewilson @phuongmy ?

PR #1527, opened 2017-04-12
ssm page external references

cy:


alex-wilmer:
you can merge yours, I'm still debugging something. thanks for the heads up

PR #1529, opened 2017-04-12
⬆️ react@15.5

alex-wilmer:
introduces `propTypes` and `createClass` warnings.

https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html


shanewilson:
any problems with the upgrade?

alex-wilmer:
Hmm problems: no. But I think some of the 3rd party stuff haven't been upgraded so the warning persists. I think it's still worth merging if we ever plan on eventually upgrading to 16.

shanewilson:
should we merge this while there are no conflict?

PR #1532, opened 2017-04-12
✨ update length of git hash display to match github

Jephuff:
this allows copying the hash and using chrome to search for the hash in the commit list


alex-wilmer:
just chrome?

Jephuff:
ok, any browser :P

PR #1537, opened 2017-04-17
Fix/743 facet tweaks

cy:


modules/node_modules/@ncigdc/components/Aggregations/ExactMatchFacet.js
```diff
@@ -82,9 +82,10 @@ const ExactMatchFacet = compose(
                     ], false),
                   }}
                   dark={!!inputValue}
-                  onClick={() => setInputValue('')}
```
Jephuff: Is this because the link wasn't calling the onClick? I wonder if it makes more sense to get Link to call onClicks passed in?

alex-wilmer: Links should already be doing that unless we forgot to pass props along somewhere

alex-wilmer: it was overwritten https://github.com/NCI-GDC/portal-ui/blame/relay/modules/node_modules/%40ncigdc/components/Links/InternalLink.js#L40

alex-wilmer: maybe we should call both things there?

```
onClick={() => {
  if (validAttrProps.onClick)  validAttrProps.onClick()
  if (deepLink) scrollToId(deepLink)
}
```

cy: AHA.

cy: my past mistakes catching up with me, ahhh! Fix pushed.

PR #1541, opened 2017-04-17
✨♻️ PRTL-969 allow keyboard selection with facet suggestions, add loading message

cheapsteak:
Added new HoC `withSelectableList` which is now used in QuickSearch, HomeSearch, SuggestionFacet, FacetSelection, and can be used anywhere where keyboard navigation on a list is desired

Also added loading message to SuggestionFacet so it'll say Loading when there are no items and is waiting for a request to come back.

### DIFF
--- a/modules/node_modules/@ncigdc/utils/namespace.js
+++ b/modules/node_modules/@ncigdc/utils/namespace.js
@@ -4,9 +4,9 @@
 import { compose, withProps, mapProps } from 'recompose';
 
 const namespace = (ns, ...hocs) => compose(
-  withProps(props => ({ [Symbol.for('parentProps')]: props })),
+  withProps(props => ({ $parentProps: props })),
   ...hocs,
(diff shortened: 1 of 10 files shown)


modules/node_modules/@ncigdc/utils/namespace.js
```diff
@@ -4,9 +4,9 @@
 import { compose, withProps, mapProps } from 'recompose';
 
 const namespace = (ns, ...hocs) => compose(
-  withProps(props => ({ [Symbol.for('parentProps')]: props })),
+  withProps(props => ({ $parentProps: props })),
```
alex-wilmer: why not Symbol?

cheapsteak: It was errorring out for reasons unknown, figured the chances of a prop name collision on $parentProps was likely low so decided to go back to normal field name

PR #1542, opened 2017-04-18
:sparkles: PRTL-1061 MutatedGenesTable annotations col

cy:
- display is_cancer_gene_census as COSMIC icon
screen shot 2017-04-18 at 11 52 16 am


Jephuff:
maybe the icon column should be centred?

cy:
@Jephuff screen shot 2017-04-19 at 11 23 59 am

modules/node_modules/@ncigdc/containers/FrequentlyMutatedGenesTable.js
```diff
@@ -145,6 +146,11 @@ const FrequentlyMutatedGenesTableComponent = compose(
             ),
           },
           {
+            key: 'annotations',
+            title: 'Annotations',
+            style: { padding: '5px 0 0 0', textAlign: 'center' },
```
Jephuff: sorry, didn't notice this before. I think this padding should be applied to a wrapper around the icons, this style is applied to the header as well so it makes it not line up with the rest :)

cheapsteak: image

cheapsteak: wishlist for tables (unrelated to PR) @alex-wilmer: 
something like `headingStyle`, `cellStyle`

shanewilson: can't you just pass a component to the table? with styles?

alex-wilmer: 💯   I would be in favor of table redesign from the ground up

cheapsteak: Can we reference implementation from [react-bootstrap-table](https://github.com/AllenFang/react-bootstrap-table)? Their API mostly seems really elegant and sensible

alex-wilmer: sure

alex-wilmer: @shanewilson  yea we can. we had some tables in the beginning that didn't any styling, so the styles were just handled at the uikit level, but things grew out of control

cy: thoughts on this https://github.com/NCI-GDC/portal-ui/pull/1542/commits/d13565bd78ea4feeb0f7f8e679d058604b10d445 as a stop gap fix before table redesign. the `padding: '5px 0 0 0'` before was because `text-align: right` on # Mutations th made that title too close to `Annotations`

cy:
@cheapsteak fixed

PR #1550, opened 2017-04-18
🐛💄PRTL-849 bandaid fix for overlapping primary sites in githut

cheapsteak:


modules/node_modules/@ncigdc/components/GitHut.js
```diff
@@ -60,11 +77,20 @@ class GitHut extends React.Component {
...
           self.angularController = controller;
+
+          // NOTE: bandaid solution to hide duplicate primary sites created due to projects now having multiple primary sites
+          $timeout(() => {
+            const duplicatePrimarySiteTexts = _.flatten(Object.values(_.groupBy(element.find('.githut #pc svg #labels .label:last-child text'), 'innerHTML')).filter(els => els.length > 1).map(_.tail));
+            duplicatePrimarySiteTexts.forEach(text => {
+              text.style.fillOpacity = 0;
+              text.style.pointerEvents = 'none';
+            });
+          }, 3000);
```
Jephuff: 3 seconds seems long, any way we can shorten it?

cheapsteak: Not without trading off on how certain we are that the graph is done rendering, or ideally figuring out an alternative to $timeout (I did give up fairly quickly (under the rationale that the consequence is that a few words would look bolder for a few seconds longer than ideal))

PR #1552, opened 2017-04-19
🐛 [PRTL-1052] fix cancer dist counts

Jephuff:


alex-wilmer:
not prefixing `cases.` to available variation data was working before right? why is `cases` needed now?

Jephuff:
It works as long as the filter is only used on the case index, there were a couple spots where it was being passed to other indexes so I changed it everywhere to be consistent.

PR #1555, opened 2017-04-19
PRTL-966 Entity autocomplete update

SidSyrus:
Autocomplete result through entity endpoint instead of query endpoint

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -669,6 +669,16 @@
                 {
                   "defaultValue": null,
                   "description": null,
+                  "name": "query",
+                  "type": {
+                    "kind": "SCALAR",
+                    "name": "String",
+                    "ofType": null
+                  }
+                },
+                {
+                  "defaultValue": null,
+                  "description": null,
                   "name": "offset",
                   "type": {
                     "kind": "SCALAR",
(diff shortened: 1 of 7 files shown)


shanewilson:
this needs to be rebased - but we should probably get it in to test

cy:
depends on https://github.com/NCI-GDC/gdcapi/pull/761? Could you rebase that branch too

SidSyrus:
@NCI-GDC/dev Ready for review

cy:
is 'filters' params maybe taking precedence over 'query'? Because I still see this
screen shot 2017-05-05 at 4 18 41 pm
```
query q($filters_0: FiltersArgument) {
  projects {
    hits(query: "TCGA-" filters: $filters_0) {
      edges {
        node {
          project_id
        }
      }
    }
  }
}
```
```
{
  "filters_0": {"op": "and", "content": [{"op": "in", "content": {"field": "projects.primary_site", "value": ["Blood"]}}]}
}
```
```
{
  "data": {
    "projects": {
      "hits": {
        "edges": [
          {
            "node": {
              "project_id": "TARGET-AML"
            }
          },
          {
            "node": {
              "project_id": "TARGET-ALL-P2"
            }
          },
          {
            "node": {
              "project_id": "TARGET-ALL-P1"
            }
          }
        ]
      }
    }
  }
}
```

sidenote, commas not needed in the graphql params list

shanewilson:
this might be a bug
the query doesn't handle `should/or` queries properly.
Requires basically a rewrite like I'm doing with aggregations cause the old so super old and doesn't nest things properly.

SidSyrus:
Just adding on the note of what Shane said. Because of this bug we can't really show the ideal result for the search. Therefore if the search query has a positive match, it will order it to the top of the result list otherwise it will show the default list.

cy:
super non-ideal, front end filter the list? 😇

shanewilson:
this is a pretty critical API bug that still exists

PR #1543, opened 2017-04-18
🐛 [PRTL-1077] correct counts in gene table

alex-wilmer:

PR #1558, opened 2017-04-19
🐛 PRTL-1079 use graphql filtered case totals to display total unique cases in stacked graph title

cheapsteak:
image


alex-wilmer:
minor overlap with https://github.com/NCI-GDC/portal-ui/pull/1543 though it looks like the same change in project breakdown minus grabbing the filter from a different module, which I agree is a good idea.

PR #1567, opened 2017-04-20
Enable cohort links

Jephuff:


Jephuff:
@alex-wilmer Do those last commits need to be on this branch? They should probably be merged to relay to make it easier to maintain this branch (except the changes in the header of course)

alex-wilmer:
I was just thinking about that.. really they should be checked off of relay in their own branch with the "enabling commit" at the top so it can be tested on its own. 

once this gets the 👍  from phuong-my i'll gently put these into relay and rebase this branch

alex-wilmer:
https://github.com/NCI-GDC/portal-ui/pull/1634 when that gets merged to relay, will update this branch

PR #1568, opened 2017-04-20
🐛 fix toggle callback in ToggleableUI

alex-wilmer:
jsx/a11y complains about li not being semantically clickable.. wrap the `li` contents with something that is?


cheapsteak:
We may need to update our version of eslint-plugin-jsx-a11y, but based on latest readme, adding `role` to the element should satisfy compliance
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md

PR #1576, opened 2017-04-21
🐛 quick fix to get all occurrences for oncogrid

Jephuff:
filter by case_ids and some simple chunking if it's over 1000

### DIFF
--- a/modules/node_modules/@ncigdc/components/Oncogrid/getQueries.js
+++ b/modules/node_modules/@ncigdc/components/Oncogrid/getQueries.js
@@ -65,7 +65,7 @@ async function getOccurences(args: {
 
   if (!geneIds.length || !caseIds.length) return { data: { hits } };
   if (from + OCCURRENCE_CHUNK > 10000) {
-    console.warn('to many occurrences to fetch');
+    console.warn('too many occurrences to fetch');
     return { data: { hits } };
   }
 
--- a/modules/node_modules/@ncigdc/components/Oncogrid/getQueries.js
+++ b/modules/node_modules/@ncigdc/components/Oncogrid/getQueries.js
@@ -1,17 +1,15 @@
 /* @flow */
-/* global API */
 import { replaceFilters } from '@ncigdc/utils/filters';
 import memoize from 'memoizee';
+import { fetchApi } from '@ncigdc/utils/ajax';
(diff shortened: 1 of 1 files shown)


modules/node_modules/@ncigdc/components/Oncogrid/getQueries.js
```diff
@@ -38,34 +36,44 @@ async function getGenes({
...
+    consequenceTypes,
+    impacts,
+    currentFilters,
+    from = 0,
+    hits = [],
+  } = args;
+
+  if (!geneIds.length || !caseIds.length) return { data: { hits } };
+  if (from + OCCURRENCE_CHUNK > 10000) {
+    console.warn('to many occurrences to fetch');
```
alex-wilmer: "too many"

alex-wilmer:
This seems reasonable. What's a good way to test it?

modules/node_modules/@ncigdc/components/Oncogrid/getQueries.js
Jephuff: oops!

Jephuff:
Good question, only test I can think of is to make sure the 50 genes make sense (before this they were getting filtered in the front end so it wasn't showing 50 genes)

shanewilson:
status?

Jephuff:
needs review

alex-wilmer:
I've counted a number of times and I'm getting 49

alex-wilmer:
image

Jephuff:
hmm exploration?

alex-wilmer:
yes with no filters. should be I checking a different page?

Jephuff:
nope, just want to know where to debug :D

alex-wilmer:
I changed it from `size=50` to 51 and I get 50 now

shanewilson:
might be a bug in API

Jephuff:
ahh, I wasn't thinking. before this change it was dropping genes because the occurrences were truncated, now it's dropping them because the cases are truncated.

alex-wilmer:
> ahh, I wasn't thinking

I doubt it. Sooo not a UI issue then?

Jephuff:
yep, not a ui issue

PR #1578, opened 2017-04-21
🐛PRTL-1109 - case aggregation returns no data when term selected after first added

cheapsteak:
Should fix issue of "Custom Facet returns no data when term selected after first added"

was not able to add test due to `TypeError: (0 , _withTooltip2.default) is not a function`


cy:
sorry, missed your comment about this in the slack channel, but you should be able to mock withTooltip in the tests with something like `jest.mock('@ncigdc/uikit/Tooltip', () => {});`

PR #1584, opened 2017-04-24
🐛 only append hash to url once

Jephuff:
It was appending another hash for every retry so you'd end up with
`url?hash=value?hash=value`

### DIFF
--- a/modules/node_modules/webpack_entry/Root.js
+++ b/modules/node_modules/webpack_entry/Root.js
@@ -48,14 +48,13 @@ if (!Relay.Store._storeData._networkLayer._implementation) {
       // Add hash id to request
       next => req => {
         const [url, search = ''] = req.url.split('?');
-        const query = {
-          ...parse(search),
-          hash: md5([
-            req.relayReqObj._printedQuery.text,
-            JSON.stringify(req.relayReqObj._printedQuery.variables),
-          ].join(':')),
-        };
-        return next({ ...req, url: `${url}?${stringify(query)}` });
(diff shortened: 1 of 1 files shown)


shanewilson:
This looks like a problem right?

```
fetchTimeout: 15000,
retryDelays: (attempt) => Math.pow(2, attempt + 4) * 100, // or simple array [3200, 6400, 12800, 25600, 51200, 102400, 204800, 409600],
```

timeout is so high compared to first retry?

modules/node_modules/webpack_entry/Root.js
```diff
@@ -47,9 +47,15 @@ if (!Relay.Store._storeData._networkLayer._implementation) {
...
-        return next(req);
+        const [url, search = ''] = req.url.split('?');
+        const query = {
+          ...parse(search),
+          hash: md5([
+            req.relayReqObj._printedQuery.text,
+            JSON.stringify(req.relayReqObj._printedQuery.variables),
+          ].join(':')),
+        };
+        return next({ ...req, url: `${url}?${stringify(query)}` });
```
shanewilson: this is good.

but for simplicity/speed it is probably fine to just add the hash and ignore the other search params ( since there shouldn't be any) and mutate the req instead of spreading.

Jephuff: sounds good, I'll update

Jephuff:
re: timeout
I don't know how `retryMiddleware` handles it, but my guess is that the retry delay would be the delay after the timeout so they don't have to match up.

PR #1587, opened 2017-04-24
🔥 remove hot module loading

Jephuff:
~WIP~
requires a change to knit https://github.com/knitjs/knit/pull/55


Jephuff:
https://github.com/knitjs/knit/pull/55 should probably be merged with this change, do you see any problems with it?

Jephuff:
`babel-polyfill` shouldn't need to go after react anymore, so probably moving it to the top makes sense.

PR #1590, opened 2017-04-25
🐛 [PRTL-1107] use react-lolliplot/dist/lib

Jephuff:
~WIP~
the umd build is broken in IE, using `/dist/lib` directly


Jephuff:
Not sure, lolliplot is using a very old version of knit, so it might not be in the latest version :)

PR #1595, opened 2017-04-26
:bug: PRTL-1045 cohort pie charts filters self

cy:
Ticket was 'Cohort Pie Charts do not refresh', got a bit tripped up because thought the filters weren't being passed to the aggregations. They are, but https://github.com/NCI-GDC/gdcapi/blob/master/gdcapi/esutils/request.py#L196 aggregations don't filter themselves unless they are range, so it looked like the `gender` pie chart didn't update when `gender` was selected for example. Fixed by front end filtering per pie chart.


cheapsteak:
image
Gender pie still does not seem to respond to filters (tested on deploybot http://deploybot.gdc.cheapsteak.net:33124/exploration?filters=~%28op~%27and~content~%28~%28op~%27in~content~%28field~%27cases.demographic.gender~value~%28~%27male%29%29%29~%28op~%27in~content~%28field~%27cases.project.project_id~value~%28~%27TCGA-BRCA%29%29%29%29%29)

cy:
wrong port yo, `:33123`

PR #1598, opened 2017-04-28
🔥 no annotation id half measure

alex-wilmer:
We ask for annotation ids such that we can link to the annotations list page, filtered. For files specifically, we need a better solution (filter annotations by file somehow).

This solves relay trying to be smart and asking for missing annotation ids on nested case/file nodes


shanewilson:
while we are here:
https://github.com/NCI-GDC/portal-ui/pull/1598/files#diff-be4ceba2f47ea6229c19048fcc61af6cR157

this data already exists on the case

PR #1608, opened 2017-05-02
✨ QoL: add flag to hide mutations count

cheapsteak:
To disable mutation count requests, set `disableMutationsCount` in localStorage to truthy


alex-wilmer:
I'm okay with it. why not just `if (localStorage.disableMutationsCount) {` ?

PR #1610, opened 2017-05-03
add npm script to clear cache

cheapsteak:
note: the schema is for relay's babel transform (i think)

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   },
   "scripts": {
     "knit": "knit",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -79,6 +79,7 @@
   },
   "scripts": {
     "knit": "knit",
+    "clear": "rm -rf node_modules && yarn && knit -- schema",
```
Jephuff: maybe `clear-cache` or `clear-babel-cache` or `schema`?

cheapsteak: reinstall?

PR #1626, opened 2017-05-05
PRTL-1159 gene id -> gene symbol

cy:


modules/node_modules/@ncigdc/components/Aggregations/index.js
```diff
@@ -51,3 +51,20 @@ export const CheckedLink = styled(Link, ({
...
+  return map[field] || field;
+};
+
+export const facetValueDisplayMapper = (field, value) => {
+  const map = {
+    'genes.is_cancer_gene_census': { 1: 'Cancer Gene Census' }, // TODO: 🙏 this needs to change in the data
+  };
+
+  const innerMap = map[field];
+  return innerMap ? map[field][value] || value : value;
```
alex-wilmer: `innerMap && innerMap[value] || value;` would work I think

this is probably fine but I can imagine a scenario where this would break.. if innerMap was not an object but a string.. and value was a string property.

cy:
got rid of `facetValueDisplayMapper`, re-ran schema: depends on key_as_string from https://github.com/NCI-GDC/gdcapi/pull/776 but phuong-my says to put this on hold ✋

shanewilson:
so one idea might be to make the gene symbol a relay container like the mutations? and you pass in a gene_id and it make a query to get the corresponding gene symbol. Would make it easier to reuse where ever we need it - since it sounds like people want the convert gene_id -> symbol everywhere.

cy:
ok, I changed my mind; still using `key_as_string` because `key` still returned a string, but the string "1" 💩! So transforming value into displayValue inside the inCurrentFilters test instead. Ready for review. @NCI-GDC/oicrdevs

alex-wilmer:
It's pretty weird that there's a delay where the gene_id flickers before the symbol shows up. maybe just show a loading spinner instead of the id?

cy:
I agree, was using GreyBox before but gene_id is what phuong-my wants (it's what's used on icgc, same flicker)

alex-wilmer:
I don't think that's very good UX. I wasn't expecting it, and it's completely dependent on latency. I'm hesitant to approve this for only that reason, otherwise its good. Can we get another opinion on this?

shanewilson:
ICGC example:
https://dcc.icgc.org/search?filters=%7B%22gene%22:%7B%22id%22:%7B%22is%22:%5B%22ENSG00000133703%22,%22ENSG00000141510%22,%22ENSG00000220635%22%5D%7D%7D%7D

Can't say I like it

shanewilson:
I think I would rather have the grey box vs spinner though - with Gene Id as fallback if the request fails or finds nothing

alex-wilmer:
That sounds reasonable to me, I say spinner in a very loose sense.

cy:
gene id in a blink tag, haha. let's wait for her opinion, don't want to be grilled for bringing back the grey box.

cy:
ready for re-re-review @NCI-GDC/oicrdevs

PR #1629, opened 2017-05-05
🐛 PRTL-1109 facet shows no data

cheapsteak:
Fixes problem with custom facets showing no data after being selected

### DIFF
--- a/modules/node_modules/@ncigdc/containers/CaseAggregations.js
+++ b/modules/node_modules/@ncigdc/containers/CaseAggregations.js
@@ -203,6 +203,7 @@ export const CaseAggregationsQuery = {
   prepareVariables: (prevVariables) => _.mapValues(prevVariables, (value, key) => (
     prevVariables.shouldRequestAllAggregations
       || initialCaseAggregationsVariables[key]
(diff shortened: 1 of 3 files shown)


shanewilson:
at some point it might be worth looking into using `facets` instead of `aggregations`

cheapsteak:
Do have documentation & sample queries? How much effort would it be?

cheapsteak:
Wonder if we could do it a piece at a time or would everything need to migrate over at once?

shanewilson:
they both exists right now - just need to use it where it makes sense. the query would be:

```
query Q($facets: JSONString, $filters: FiltersArgument) {
  repository {
    cases {
      facets(facets: $facets, filters: $filters)
    }
  }
}

variables: {
 facets: ['primary_site', 'project.name'],
 filters: { ...blah...}
}
```

shanewilson:
should we merge this is a fix until we move the `facets`?

shanewilson:
https://jira.opensciencedatacloud.org/browse/PRTL-1109 says it is merged?

cheapsteak:
Yeah let's merge, it's not as good as using facets but net positive on bug reduction. 

I merged that ticket when I merged the ticket that I thought fixed it but turns out it wasn't really fixed

PR #1631, opened 2017-05-05
PRTL-1000 add download button to mac chart

cy:
~~Should we filter out `"__dataID__":`s from the 'Download JSON'? The other viz JSON downloads have them in too.~~


cy:
change of approach, the CancerDistribution chart always exported chart data as `{ label, value }` but the other charts exported the raw data with data_ids, scores etc. Consistently always export `{ label, value }` instead. Ready for review.

PR #1632, opened 2017-05-08
🎛PRTL-1162 Remove analysis menu from menu bar

cheapsteak:
image


shanewilson:
does this feature need to be turned on and off?

cheapsteak:
According to the ticket, "We'll add it back when we show analyses (next DAVE)."

PR #1640, opened 2017-05-09
✨ [PRTL-1166] add consequence type filter to lolliplot

Jephuff:
Also created a filter for the consequence types so it can be shared.
depends on https://github.com/NCI-GDC/gdcapi/pull/779


modules/node_modules/@ncigdc/utils/filters/prepared/significantConsequences.js
```diff
@@ -0,0 +1,23 @@
+// @flow
```
cheapsteak: woot, another prepared filter :D

alex-wilmer:
lgtm

image

PR #1642, opened 2017-05-09
:bug: PRTL-1176 keep searchTableTab after CurrentFilters clear

cy:


modules/node_modules/@ncigdc/components/CurrentFilters.js
```diff
@@ -130,7 +132,7 @@ const CurrentFilters = ({
         <Row wrap spacing="0.3rem">
           <NotUnderlinedLink
             style={styles.groupPadding}
-            query={{}}
+            query={{ searchTableTab }}
```
Jephuff: It should probably keep other things like the facet tab as well.
Might be better to specify the things we are clearing instead? either by passing in the whole query and omitting filters or making `query={{filters: null}}` work.

cy:
@Jephuff ready for re-review

modules/node_modules/@ncigdc/components/CurrentFilters.js
```diff
@@ -57,6 +58,7 @@ const LinkButton = styled(Link, {
 });
 
 type TProps = {
+  searchTableTab: string,
```
Jephuff: this isn't used any more is it?

cy: opps, updated

PR #1645, opened 2017-05-09
✨ [PRTL-1156] update truncating lengths

Jephuff:


Jephuff:
I'm always conflicted. It changes behavior so if someone was mousing over them one at a time and then one didn't have a tooltip it might be unexpected... but really... it is silly hahaha
Should I remove it in that case?

PR #1646, opened 2017-05-09
🛑 🎛 (requested to be blocked) PRTL-1174 switch links that use gene id in filter to use gene symbol

cheapsteak:


modules/node_modules/@ncigdc/containers/CancerDistributionTable.js
```diff
@@ -99,7 +98,7 @@ const CancerDistributionTableComponent = compose(
           key={b.key}
           ssms={explore.ssms}
           filters={makeFilter([
-            { field: 'genes.gene_id', value: [geneId] },
+            { field: 'genes.symbol', value: [entityName] },
```
cy: I'm not sure that this change, it's the filter for fetching mutation counts for the gene in each project, not a link to the explore page and junjun's said there could be genes with multiple ids? Might change the counts.

cy: correcting that, "it's the two genes have the same symbol, the symbol is marked with a `*`"

cy: same with ProjectBreakdown and the other table filters, I think should only change the `ExploreLinks` and the filters on `push` 🤔

cheapsteak:
PR has been blocked
https://oicr-gdc.slack.com/archives/C0H0726P7/p1494435092757308
>Just had a discussion with Junjun on the Gene symbol / Gene id search, it does not work.
>Can you please hold on PRTL-1174 (@chang) and PRTL-1159

PR #1651, opened 2017-05-10
✨ retry middleware that takes the first successfull query

Jephuff:
Took a stab at making the retry middleware take the first successful response, seems to work.
Needs more testing/refactoring but thought I would be it here if anyone wants to take a look.

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -7,11 +7,8 @@ import Router from "react-router-dom/BrowserRouter";
 import { stringify, parse } from "query-string";
 import md5 from "blueimp-md5";
 import urlJoin from "url-join";
-import {
-  RelayNetworkLayer,
-  urlMiddleware,
-  retryMiddleware
-} from "react-relay-network-layer";
+import { RelayNetworkLayer, urlMiddleware } from "react-relay-network-layer";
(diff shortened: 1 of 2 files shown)


shanewilson:
what did you change?

Jephuff:
mainly timeouts don't reject the promise and creates a new promise thats put in an array with a promise.race

shanewilson:
can you rebase?

shanewilson:
what page are you testing to see reduced requests?

Jephuff:
connected to UAT on the exploration page it would have multiple requests, it's the only consistent case I've found

Jephuff:
Had to use this branch to be able to click the facets with data checkbox, probably worth looking at merging this after today.

shanewilson:
what happened?

Jephuff:
The requests time out and max retries fire. With this retries fire until the first request returns. I guess we could also increase timeout.

shanewilson:
how did you reproduce this? it would be nice to see it in action

Jephuff:
hmmm, now it's loading quite quickly, I wonder if something was happening to the UAT server.
all I did before, was running api locally point to UAT go to explore page and 
`Add a Case Filter` -> `Only show fields with values`
I'll keep my eye out for another case

PR #1661, opened 2017-05-11
✨ PRTL-1133 Gene and Mutation bar graphs should have different behaviour based on context (Project or Exploration)

cheapsteak:
https://jira.opensciencedatacloud.org/browse/PRTL-1133

### DIFF
--- a/modules/node_modules/@ncigdc/components/Explore/GenesTab.js
+++ b/modules/node_modules/@ncigdc/components/Explore/GenesTab.js
@@ -1,13 +1,15 @@
 /* global API */
 // @flow
 import React from 'react';
-import { compose, withState, withProps, withPropsOnChange } from 'recompose';
+import JSURL from 'jsurl';
+import { compose, withState, withProps, withPropsOnChange, withHandlers } from 'recompose';
 import { Row, Column } from '@ncigdc/uikit/Flex';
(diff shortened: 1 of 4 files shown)


alex-wilmer:
So I did a super quick test to see if this would work. 

```
import React from 'react'
import { BrowserRouter as R, Route, Link } from 'react-router-dom'

let Chart = () =>
  <div>
    <h1>chart</h1>
    <svg>
      <Link to='/a'>
        <circle cx={50} cy={50} r={10} fill='red' />
      </Link>
    </svg>
  </div>


export default () =>
  <R>
    <div>
      <Route exact path='/' component={Chart} />
      <Route path='/a' component={() => <div>page</div>} />
    </div>
  </R>

```

and it does. not that this invalidates your PR, but I'm thinking it might be worth it to drop d3 for simple things like bar charts so we don't have to stack imperative stuff on top of each other like `push`

cheapsteak:
I am onboard with declarative component-based svg graphs, even for complex graphs as well (maybe especially complex graphs)

I don't see a way to sell rewriting what we already have, but for new graphs it'd be cool to experiment.

d3 may still be useful though for utility functions like `lineGenerator` (This article is pretty cool  https://medium.com/@ilikepiecharts/playing-with-d3-version-4-react-react-motion-3d04c6eb21c9)

In the meantime, approve or request changes?

PR #1666, opened 2017-05-12
:sparkles: PRTL-1187 export html table as tsv

cy:
~~[WIP]~~ Couldn't think of a better way to do this with `<ProjectBreakdown>` and `<MutationsCount>` both being self-contained relay Containers that make secondary calls. I'm not sure how to get the data out of them in the parent. Straight up using document.querySelector. Taking suggestions for better solutions.

done, ~~pending specs confirmation~~ also done!

### DIFF
--- a/modules/node_modules/@ncigdc/components/DownloadTableToTsvButton.js
+++ b/modules/node_modules/@ncigdc/components/DownloadTableToTsvButton.js
@@ -38,7 +38,7 @@ const DownloadTableToTsvButton = ({
           return reduce(tds, (acc, td) => {
             const markedForTsv = td.querySelector('.for-tsv-export');
             const exportText = markedForTsv ? markedForTsv.innerText : td.innerText;
-            const joinedText = exportText.trim().replace(/\s*/, ' ').split('\n').join(',');
+            const joinedText = exportText.trim().replace(/\u00A0*/, ' ').split('\n').join(',');
             const colspan = td.getAttribute('colspan');
             const fittedToColspan = colspan ? [joinedText, ...Array(colspan - 1)] : [joinedText];
             return [...acc, ...fittedToColspan];
--- a/modules/node_modules/@ncigdc/components/DownloadTableToTsvButton.js
+++ b/modules/node_modules/@ncigdc/components/DownloadTableToTsvButton.js
@@ -38,7 +38,7 @@ const DownloadTableToTsvButton = ({
           return reduce(tds, (acc, td) => {
             const markedForTsv = td.querySelector('.for-tsv-export');
             const exportText = markedForTsv ? markedForTsv.innerText : td.innerText;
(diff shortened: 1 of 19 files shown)


cy:
this passed specs review and ready for code review @NCI-GDC/oicrdevs. Finicky things:

- can't use `&nbsp;` because Excel refuses to respect UTF8 encoding and shows them as `Â`, if the encoding is set then it'll also show UTF8's byte order mark at the begining of the file as ï»¿ http://stackoverflow.com/a/155176/962292 :/ So don't set the encoding and replace `&nsbp;` with template string
- UUIDs are displayed truncated but are wanted exported in full, and icons are wanted to be exported as text so created `<ForTsvExport>` for these cases

modules/node_modules/@ncigdc/components/DownloadTableToTsvButton.js
```diff
@@ -0,0 +1,64 @@
...
+        const headThs = map(headTrs, h => h.querySelectorAll('th'));
+        const thEls = headThs.length === 2 ? getSingleHeader(headThs) : tableEl.querySelectorAll('th');
+        const thText = map(thEls, el => el.innerText).map(t => t.replace('\n', ' '));
+        const trs = tableEl.querySelector('tbody').querySelectorAll('tr');
+        const tdText = map(trs, t => {
+          const tds = t.querySelectorAll('td');
+          return reduce(tds, (acc, td) => {
+            const markedForTsv = td.querySelector('.for-tsv-export');
+            const exportText = markedForTsv ? markedForTsv.innerText : td.innerText;
+            const joinedText = exportText.trim().split('\n').join(',');
```
Jephuff: Instead of requiring no `&nbsp;` could we just add something like `.replace(/\s*/g, ' ')` here?

cy: good idea!

modules/node_modules/@ncigdc/components/DownloadTableToTsvButton.js
```diff
@@ -38,7 +38,7 @@ const DownloadTableToTsvButton = ({
           return reduce(tds, (acc, td) => {
             const markedForTsv = td.querySelector('.for-tsv-export');
             const exportText = markedForTsv ? markedForTsv.innerText : td.innerText;
-            const joinedText = exportText.trim().replace(/\s*/, ' ').split('\n').join(',');
```
Jephuff: it should be catching the `&nbsp;` there is likely just another whitespace character before it :D just needs the global flag on the regex :)

cy: did both cuz of this answer http://stackoverflow.com/a/11819337/962292

Jephuff: Makes sense 👍

PR #1669, opened 2017-05-15
🐛 [PRTL-1206] fix duplicates in oncogrid

Jephuff:
~WIP~
remove duplicate mutations from the oncogrid
Also reduces time to map occurrences from about 2700ms to 160ms

### DIFF
--- a/modules/node_modules/@ncigdc/components/Oncogrid/dataMapping.js
+++ b/modules/node_modules/@ncigdc/components/Oncogrid/dataMapping.js
@@ -48,7 +48,7 @@ export const mapDonors = (donors: Array<TDonorInput>): Array<TDonor> => (
     const output = { ...dataTypesInitial };
 
     summary.data_categories.forEach((category) => {
-      output[category.data_category] = category.file_count || 0;
+      output[category.data_category] = category.file_count || 0; // eslint-disable-line fp/no-mutation
     });
 
     return {
@@ -115,9 +115,10 @@ export type TOccurence = {
   functionalImpact: string,
 };
 
-function expandObs(geneIdToSymbol) {
+function expandObs(geneIdToSymbol: {}): Function {
   return ({ ssm, case: casObj = {} }): Array<TOccurence> => ssm.consequence
     .map(c => c.transcript)
+    .filter(({ annotation = {} }) => annotation.impact)
     .map((consequence): TOccurence => {
       const { annotation = {}, gene = {}, consequence_type } = consequence;
       const geneId = gene.gene_id;
@@ -136,13 +137,15 @@ function expandObs(geneIdToSymbol) {
     });
 }
 
-export const mapOccurences = (
+type TMapOccurences = (
   occurrences: Array<TOccurenceInput>,
   donors: Array<TDonor>,
   genes: Array<TGene>,
   consequenceTypes: Array<string>,
   impacts: Array<string>
-): Array<TOccurence> => {
+) => Array<TOccurence>;
+
+export const mapOccurences: TMapOccurences = (occurrences, donors, genes, consequenceTypes, impacts) => {
   const donorIds: Array<string> = donors.map((d: TDonor): string => d.id);
   const geneIds: Array<string> = genes.map((g: TGene): string => g.id);
   const geneIdToSymbol = genes.reduce((acc, g: TGene) => ({ ...acc, [g.id]: g.symbol }), {});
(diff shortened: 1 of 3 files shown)


modules/node_modules/@ncigdc/components/Oncogrid/dataMapping.js
```diff
@@ -115,50 +115,52 @@ export type TOccurence = {
...
-      ((consequenceTypes || []).includes(occurrence.consequence)) &&
-      geneIds.indexOf(occurrence.geneId) >= 0 &&
-      donorIds.indexOf(occurrence.donorId) >= 0 &&
-      (!impacts.length || impacts.includes(occurrence.functionalImpact))
-    );
+  const arr = [];
+  for (let i = 0; i < occurrences.length; i += 1) {
+    const { ssm: { consequence, ssm_id }, case: { case_id } = {} } = occurrences[i];
+
+    if (caseIds.includes(case_id)) {
```
alex-wilmer: Since you're already iterating over donors before this line, you can create a set or map instead, so that this can be a hash lookup instead of an includes iteration. I was trying to find some info about lookup time differences, but I think Set makes semantically the most sense and appears to be fast by some measure.

http://stackoverflow.com/questions/37464704/set-lookup-time-slower-than-object

modules/node_modules/@ncigdc/components/Oncogrid/oncoGridParams.js
```diff
@@ -97,10 +97,17 @@ export default function ({
...
+  const donorObs = new Set();
+  const geneObs = new Set();
+  for (let i = 0; i < observations.length; i += 1) {
+    const observation = observations[i];
+    donorObs.add(observation.donorId);
+    geneObs.add(observation.geneId);
+  }
+
+  donors = donors.filter(d => donorObs.has(d.id)); // eslint-disable-line fp/no-mutation
+  genes = genes.filter(g => geneObs.has(g.id)); // eslint-disable-line fp/no-mutation
```
shanewilson: could you do this the other way? in the observation loop check if that `donor.id` is in `donors` and then make the set in the loop.

map(Donors/Genes) should probably return an Object of `id: object`

then you don't need to make the `(d/g)Obs` sets and you don't need to loop/filter over `donors/genes`.

the only loop you make is over observation.

shanewilson: well you still need to make two sets but they end up being the final sets that you use in the rest of the function rather than those temp sets.

Jephuff: I started doing something similar and then bailed when it wasn't improving performance haha
I'll take another look at it :)

shanewilson: well donors and genes are pretty small compared to the other lists but it doesn't add complexity to the code does it?

still seems clear that you are looking for intersection of donor/gene ids

shanewilson: if it makes the code harder to understand or use probably not worth it

Jephuff: how do you feel about something like this 
```const { observations, donorIds, geneIds } = buildOccurences(occurencesData, donorData, geneData, consequenceTypes, impacts);
if (observations.length === 0) return null;
const donors: Array<TDonor> = mapDonors(donorData, donorIds);
const genes: Array<TGene> = mapGenes(geneData, geneIds);
```
then `mapGenes` and `mapDonors` filter themselves

shanewilson: so basically:

```

  const observations: Array<TOccurence> = mapOccurences(occurencesData, donors, genes, consequenceTypes, impacts);

  if (observations.length === 0) return null;

  // Clean gene & donor data before using for oncogrid.

  const donorObs = new Set();
  const geneObs = new Set();
  for (let i = 0; i < observations.length; i += 1) {
    const observation = observations[i];
    donorObs.add(observation.donorId);
    geneObs.add(observation.geneId);
  }
  let donors: Array<TDonor> = mapDonors(donorData, donorObs); // eslint-disable-line fp/no-let
  let genes: Array<TGene> = mapGenes(geneData, geneObs); // eslint-disable-line fp/no-let
```

?

sure up to you.

Jephuff: yep, but `geneObs` and `donorObs` are built inside the existing occurrence loop. so maybe a bit faster if there are a lot of occurrences? probably not noticeable though haha

PR #1673, opened 2017-05-16
✨ arrange columns for projects table

alex-wilmer:


alex-wilmer:
This took some fussing around.. but I think it's all working now. This is currently just for the projects list table but just a matter of copy / pasting to get it in the other main tables. This one has the most complexity (subheadings + totals in footer) out of the rest.

modules/node_modules/@ncigdc/containers/ProjectsTable.js
```diff
@@ -0,0 +1,102 @@
...
+  return (
+    <div>
+      <Row style={{ backgroundColor: 'white', padding: '1rem', justifyContent: 'space-between', alignItems: 'center' }}>
+        { tableHeader &&
+          <h3 className='panel-title'>
+            {tableHeader}
+          </h3>
+        }
+        <Showing
+          docType={entityType}
```
cy: should prob. standardize whether this is called `docType` or `entityType` https://github.com/NCI-GDC/portal-ui/pull/1673/files#diff-3bd0128dc189917fed9943962383b4a6R37

shanewilson: docType should probably stay an API only thing? we don't want to show stuff like `gene-centric` or whatever on the UI

cheapsteak:
👍  huge improvement

some subjective api suggestions (feel free to ignore)

```js
const columnDefinition = {
  name: 'ID',
  id: 'project_id',


  // heading would be optional and have a default of
  // ({columnDefinition}) => <Th key="{columnDefinition.id}">columnDefinition.name</Th>,
  heading: ({ columnDefinition }) => <Th key="{columnDefinition.id}" rowSpan="2">columnDefinition.name</Th>,

  sortable: true,
  
  // the raw value would be useful for csv/tsv exports
  cellValue: ({ row, tableData }) => row.node.project_id,

  // the passed-in cellValue would be the result of running the `cellValue` function using data for that row
  renderCell: ({ cellValue, columnDefinition, row, tableData }) => (
    <Td key={columnDefinition.id}>
      <ProjectLink>
        {cellValue}
      </ProjectLink>
    </Td>
  ),
},
```

cheapsteak:
I remember we also at some point talked about making the columns be compose-able, since we have tables that are similar but slightly different

This would allow to instead of have 

```js
<ProjectsFrequentlyMutatedGenesTable/>
<ExplorationFrequentlyMutatedGenesTable/>
```

We could have 

```js
<Table>
  {commonColumns}
  <UniqueColumn foo={"something"}>
</Table>

<Table/>
  {commonColumns}
  <UniqueColumn foo={"something else"}>
</Table>
```

Here's an example
https://codepen.io/cheapsteak/pen/pPOdvj?editors=1010

modules/node_modules/@ncigdc/containers/ProjectsTable.js
cy: so we should change the `docType`s to `entityType`? `docType` is also used in the facet components

alex-wilmer: there is a difference between `docType` and `entityType`. I'm not sure it makes sense to name them the same everywhere on the ui, depends on how its used. Probably a discussion to have outside the scope of this ticket. I think we can safely assume entityType is anything that might have a table on the ui (file, project, associated entity, biospecimen etc) while docType is the type in the index?

cy: sure some of it might be my misunderstanding when making the facet components

alex-wilmer:
@cheapsteak your column definition spec looks good. Also I think the extra composability is good to have. It will become more apparent when getting to the viz index tables

cheapsteak:
Looks great, 2 minor things I've noticed

- [ ] Showing 1- 39 of 39 projects
image
- [ ] ID should be Project ID

alex-wilmer:
thanks for checking. updated ID.

I removed the "showing" part since there's no pagination in this table. this is prod: https://portal.gdc.cancer.gov/projects/t

cy:
good job changing ID -> Project ID, excel thinks that csv/tsv files that start with ID are SYLK https://superuser.com/questions/210027/why-does-excel-think-csv-files-are-sylk and gives a warning

PR #1676, opened 2017-05-17
travis jobs?

shanewilson:


cheapsteak:
Turn off email notifications?
https://docs.travis-ci.com/user/notifications#Configuring-email-notifications

shanewilson:
@NCI-GDC/oicrdevs 
votes on that? ^

cheapsteak:
We can also have a list in the config for people who opt-in to email notifications

cy:
never read them, the PR status is enough

.travis.yml
```diff
@@ -7,21 +7,23 @@ cache: yarn
...
-# - yarn lint
-# - yarn flow
-- yarn test -- --no-cache
-after_success:
-- NODE_ENV=production COMMIT_HASH=$TRAVIS_COMMIT yarn build
-- docker login -e="$DOCKER_EMAIL" -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" quay.io
-- bash docker.sh
 before_install:
 - yarn --version
 - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.21.3
```
shanewilson: @Jephuff do you remember why we pinned yarn?

Jephuff: travis broke our builds awhile back because they include an old version of yarn and then updated the install command to use something that was only in newer versions. 
not sure if thats still an issue, but it still has an old version of yarn

PR #1679, opened 2017-05-17
✨ PRTL-1138 projects numcases link explore

cheapsteak:


cheapsteak:
Sorry @Jephuff got updated specs and the way filters work had to be changed again (now going back to filtering on project id)

modules/node_modules/@ncigdc/components/ProjectsCharts.js
```diff
@@ -87,28 +88,18 @@ const initialState = {
...
-    } : null,
-  ].filter(Boolean);
-
+function getGenes({ relay, caseCountFilters }: TProps): void {
   relay.setVariables({
     fetchGeneData: true,
     fmgCaseCount_filters: {
       op: 'AND',
-      content: commonFilters,
+      content: caseCountFilters,
```
Jephuff: `caseCountFilters` could be `[]` so probably should check that and make `fmgCaseCount_filters` null if it is.

modules/node_modules/@ncigdc/components/ProjectsCharts.js
```diff
@@ -305,8 +308,14 @@ const ProjectsChartsComponent = compose(
         </div>
         {!genesIsFetching ? [
           <div style={{ alignSelf: 'center', color: theme.greyScale7, fontSize: '1.2rem' }} key='bar-subtitle'>
+            <ExploreLink
+              query={{
+                searchTableTab: 'cases',
+                filters: { op: 'and', content: caseCountFilters },
```
Jephuff: same as above, `caseCountFilters` could be `[]`

PR #1687, opened 2017-05-19
Feat/1208 external db facets as 'not missing' query

cy:
depends on https://github.com/NCI-GDC/gdcapi/pull/788

### DIFF
--- a/modules/node_modules/@ncigdc/utils/__tests__/string.test.js
+++ b/modules/node_modules/@ncigdc/utils/__tests__/string.test.js
@@ -1,0 +1,24 @@
+/* @flow */
+
+import { capitalize } from '../string';
+
+describe('capitalize', () => {
+  it('should captalize a string', () => {
+    expect(capitalize('danish marzipan biscuit')).toEqual('Danish Marzipan Biscuit');
+  });
+
+  it('should capitalize mirna as miRNA', () => {
+    expect(capitalize('mirna')).toEqual('miRNA');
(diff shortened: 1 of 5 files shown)


cy:
this is ready for review. Adds `NotMissingFacet` component under Mutation facet panel that sets a `not missing` query. Does not actually use aggregations for counts because ran into 100 buckets size limit, so using filtered hits total.

modules/node_modules/@ncigdc/utils/__tests__/string.test.js
```diff
@@ -0,0 +1,24 @@
...
+import { capitalize } from '../string';
+
+describe('capitalize', () => {
+  it('should captalize a string', () => {
+    expect(capitalize('danish marzipan biscuit')).toEqual('Danish Marzipan Biscuit');
+  });
+
+  it('should capitalize mirna as miRNA', () => {
+    expect(capitalize('mirna')).toEqual('miRNA');
+    expect(capitalize('danish marzipan mirna biscuit')).toEqual('Danish Marzipan miRNA Biscuit');
```
alex-wilmer: 🍪

cy: mmm miRNA biscuits

cy:
@alex-wilmer you mean these https://github.com/NCI-GDC/portal-ui/pull/1687/files#diff-f3e533b9b48faa3f3cb23696da58d47dR162 filters? Don't expect those to change, so I think it's ok having them directly in the string.

PR #1688, opened 2017-05-23
✏️ PRTL-1128 p-value copy for when 0 should still be visible

cheapsteak:


modules/node_modules/@ncigdc/components/SurvivalPlotWrapper.js
```diff
@@ -156,7 +156,7 @@ const SurvivalPlotWrapper = ({
             {
               <div className="p-value">
                 <div style={styles.pValue}>
-                  {pValue && `Log-Rank Test P-Value = ${pValue.toExponential(2)}`}
+                  {`Log-Rank Test P-Value = ${pValue.toExponential(2)}`}
```
cy: this happens 
```
Uncaught TypeError: Cannot read property 'toExponential' of undefined
    at SurvivalPlotWrapper (SurvivalPlotWrapper.js:159)
```
maybe `(pValue || 0). toExponential(2)`?

(also ticket number in branch name & commit msg don't match)

Jephuff: I don't think you can get rid of that check, with 1 curve it will be undefined and you don't want to show anything.

cheapsteak: change it to `pValue === 0`?

cheapsteak: `!_.isNil(pValue)`

Jephuff: if your using lodash, something like `isNumber` would likely be better :)

PR #1697, opened 2017-05-24
gene table renderer (fix pagination issue)

alex-wilmer:
this makes it more difficult to handle loading state *inside* the component.. maybe we don't need that


alex-wilmer:
Can someone take a look at this so we can implement this in pieces? This is only implemented currently on the project entity page, for the gene table.

disregard the misnamed branch 😄

cy:
it's better. Except does feel slower, and sometimes the mutations counts get stuck at grey box after adding a facet. Might go back to what you were saying earlier, to also make those their own renderer.

shanewilson:
it would probably be better to look into changing up those mutation queries into a single facet query.

PR #1700, opened 2017-05-25
[wip] 1219 mutations count

cy:
maybe a fix - actually couldn't reproduce before. Opening pr for BA to test.

### DIFF
--- a/modules/node_modules/@ncigdc/containers/FrequentlyMutatedGenesTable.js
+++ b/modules/node_modules/@ncigdc/containers/FrequentlyMutatedGenesTable.js
@@ -126,16 +126,16 @@ const FrequentlyMutatedGenesTableComponent = compose(
         caseTotal={g.case.hits.total}
       />
     ),
-    num_mutations: React.createElement(MutationsCount, {
-      key: g.gene_id,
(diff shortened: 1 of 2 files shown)


alex-wilmer:
interesting. why was createElement used there? I think those two things are equivalent. also I'm thinking that mutations count should be its own renderer too which would surely fix this in case this doesn't

cy:
yeah, I was testing out createElement instead of jsx one time and committed it accidentally. The thing I was thinking might make a difference was passing in query.fmgTable_filters as well. But it does look like defaultFilters is the thing used on the GenesTab table.

cy:
and Bessie's saying it was the CaseTable after all :x lol

alex-wilmer:
mmm can we leave this? I'm trying to remove all the force fetching. or maybe you want to test making mutations count into its own renderer like one of these: https://github.com/NCI-GDC/portal-ui/tree/0508f71789ef856f14205185e83a4dd5914776e0/modules/node_modules/%40ncigdc/modern_components

cy:
relayyyyyyy. fine

cy:
to be replaced

PR #1702, opened 2017-05-25
🐛✅ PRTL-1233 Range facets turn into Term facets when query changes

cheapsteak:


cy:
looks like has the strange effect of, it 2 range filters are added, the first one is lost 🤔

cheapsteak:
That happens also on relay
Merge and make new ticket?
Less merge conflicts

cy:
🎫 https://jira.opensciencedatacloud.org/browse/PRTL-1236

cheapsteak:
range filters seems to be broken currently on relay branch 
hesitant to merge this in since this could be blamed as the culprit

cy:
broken in what way, wrong data returning or filter set wrong?

cheapsteak:
Is working for me now as well, weird
Adding a range filter was resulting in a javascript error complaining something about ctx

modules/node_modules/@ncigdc/utils/filters/index.js
```diff
@@ -81,7 +81,7 @@ export const replaceFilters: TMergeFilters = (q, ctxq) => {
   const merged = {
     op: 'and',
     content: ctxq.content.reduce((acc, x) => {
```
alex-wilmer: is this `x` supposed to be `ctx`?

cheapsteak: I think you're right

cheapsteak: or the other `ctx`s should be `x`

cheapsteak: PR open
thanks for catching this
https://github.com/NCI-GDC/portal-ui/pull/1708/files

PR #1706, opened 2017-05-26
🐛 move polyfills back for IE11

Jephuff:
~Move this to the top when we remove hot reloading, looks like I was wrong and it still needs to be after react. (waiting to test on collab)~
That ~was~ wasn't it

### DIFF
--- a/modules/node_modules/webpack_entry/index.js
+++ b/modules/node_modules/webpack_entry/index.js
@@ -2,9 +2,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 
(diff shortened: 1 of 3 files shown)


Jephuff:
It wasn't it... but I did write `was it`. Sorry!

cy:
i'm so confused lol

cy:
but i did manage to get it to work in ie11 without it

Jephuff:
yeah i'm just pushing it for the record( and so I can test :D ) :) 
Does merge maintain order?

PR #1707, opened 2017-05-26
🐛 wrong var name

alex-wilmer:


alex-wilmer:
thanks, just making a better commit message and will merge

PR #1711, opened 2017-05-27
Try cra

shanewilson:


src/index.js
```diff
@@ -0,0 +1,22 @@
+/* @flow */
+
+// import "babel-polyfill";
```
Jephuff: Did you comment this out on purpose? It's broken in IE because it's missing polyfills.

shanewilson: I did - that would mean cra is broken on IE?
what features is IE missing?

Jephuff: Well cra isn't broken in IE, it just doesn't bundle the polyfills our code needs. The first one it breaks on is `Object.values` but I'm guessing there are quite a few other ones we are using.

PR #1712, opened 2017-05-28
✨ PRTL-894 cases table arrange columns

alex-wilmer:


src/packages/@ncigdc/tableModels/utils.js
```diff
@@ -0,0 +1,101 @@
...
+) => {
+  let idFilter; // because of weird ternary bug https://github.com/babel/babel/issues/5809
+
+  if (type === "case") {
+    idFilter = node => ({
+      field: "cases.project.project_id",
+      value: node.project_id
+    });
+  } else {
+    idFilter = node => ({
```
cheapsteak: Can we pass this in from the call location instead of internalizing it into the function?

cheapsteak: On second glance I would propose we add 2 additional options that take in functions

```
createDataCategorySubColumns({
  ...
  getCellLinkFilters: (rows, category) => makeFilter(...)
  getTotalLinkFilters: (rows, category) => makeFilter(...)
});
```

what do you think?

cheapsteak:
Not sure if this is a concern, but the order of the cases are different

relay: http://deploybot.gdc.cheapsteak.net:33068/repository
image


this pr: http://deploybot.gdc.cheapsteak.net:33060/repository
image

Have not set any sort order or filters

cheapsteak:
Relay shows:
"Available Files per Data Category"

This PR has
"Available Cases per Data Category"

cheapsteak:
Sort order probably doesn't matter, the number of results is the same, as is the filesize

Let's change "Available Cases" to "Available Files" and merge this

alex-wilmer:
@cheapsteak thanks. going to update the util function to have all that stuff passed in from caller

alex-wilmer:
@cheapsteak updated

PR #1718, opened 2017-05-29
🐛 use id for key instead of index entity table

alex-wilmer:


src/packages/@ncigdc/modern_components/GenesTable/GenesTable.js
```diff
@@ -191,6 +191,7 @@ const Component = compose(
           </Row>
         </Row>
         <EntityPageHorizontalTable
+          idKey="gene_id"
```
cheapsteak: Can we repurpose `key` here? seems appropriate

alex-wilmer: react doesn't want you to access `props.key`, it will throw a warning

PR #1728, opened 2017-05-30
♻️💄 PRTL-1240 project list table abbr underline, make th accept functions

cheapsteak:


Jephuff:
Did you say the functions around the Th elements were because theme wasn't initialized yet? 
Could we just move the setTheme call above the `./Root` import? or make `active` an environment variable and remove setTheme?

alex-wilmer:
Chang is saying function interchangeably with component. They were meant to be components in the first place. The fact that they weren't was just an oversight I made when copying things over from angular.

Jephuff:
ahh, that makes sense 👍

alex-wilmer:
Thanks for helping me out with this. I am just a tiny bit hesitant on looping over the categories to create the columns. It helps the code be more concise but we will have to unwind that loop in the future if a single column requires different treatment.

I don't actually think this will be the case though so this is probably better for now.

PR #1729, opened 2017-05-31
🐛 fix null crash in case entity

alex-wilmer:


alex-wilmer:
testing case: http://localhost:3000/cases/04dc8ee3-ec93-4b56-a546-f96746f8122f

src/packages/@ncigdc/containers/CasePage.js
```diff
@@ -72,11 +69,9 @@ export const CasePageComponent = ({ node, viewer }: TProps) => (
           query={ctx.query}
           node={node}
           totalFiles={node.files.hits.total}
-          ssmTested={_.get(
-            viewer.explore.cases.hits,
-            "edges[0].node.available_variation_data",
-            []
-          ).includes("ssm")}
+          ssmTested={(viewer.explore.cases.hits.edges[0].node
```
Jephuff: Is it possible edges will be empty at some point? Maybe we should still guard against that?

alex-wilmer: nope because that will fail at the route level

alex-wilmer: well.. that is to say, it's failing currently at the route level because ES returns a 404 style error

Jephuff: makes sense 👍

PR #1731, opened 2017-05-31
✨ [PRTL-1254, PRTL-1267] Remove components from project page

Jephuff:


shanewilson:
close?

Jephuff:
Not sure they are done thinking about the case tab removal?

PR #1739, opened 2017-06-01
🐛PRTL-1246 don't wrap gql filters with additional "and" if already has "and"

cheapsteak:


Jephuff:
Theres no change?

cheapsteak:
@Jephuff I accidentally pushed the commit into relay, lol

PR #1742, opened 2017-06-01
✨ [PRTL-1185] legend in lolliplot svg

alex-wilmer:


Jephuff:
The legend in the png export is broken :(
protein_viewer-braf-2017-06-02

alex-wilmer:
@Jephuff removed the flex stuff and seems to work now for both png and svg

src/packages/@ncigdc/containers/Lolliplot.js
```diff
@@ -467,10 +467,7 @@ const LolliplotComponent = compose(
                   margins: { top: 20, right: 20, bottom: 20, left: 20 },
                   embed: {
                     right: {
-                      width: 350,
-                      margins: {
-                        left: -350
```
Jephuff: May add margins back with `right: 20` the legend goes right to the edge.

alex-wilmer:
good call. I also added more margin to the left hand side of the svg which looks nicer

PR #1746, opened 2017-06-01
♻️ PRTL-1255 add active styles to nav link

cheapsteak:
Did not remove the bootstrap classes because underlines might start appearing in a bunch of places where we aren't expecting it

Added selectors to navlink items


src/packages/@ncigdc/components/Header.js
```diff
@@ -22,10 +22,15 @@ import { setModal } from "@ncigdc/dux/modal";
 import SessionExpiredModal from "@ncigdc/components/Modals/SessionExpiredModal";
 
 import Banner from "@ncigdc/uikit/Banner";
+import activeTheme from "@ncigdc/theme/versions/active";
```
Jephuff: I don't think we should be requiring the active theme directly.
but I also don't think we need to worry about theme not being initialized. 
if we move `setTheme("active");` inside theme and maybe use an env variable for the theme name then we can just export theme and directly use it.

cheapsteak: Those are great ideas, I would approve that PR
In the meantime I went with hopefully a path of less resistance

alex-wilmer:
@cheapsteak I have a small suggestion before merging one sec

src/packages/@ncigdc/components/Links/InternalLink.js
```diff
@@ -27,7 +38,10 @@ const InternalLink = ({
 
   const q = removeEmptyKeys ? removeEmptyKeys(q1) : q1;
 
-  const validAttrProps = validAttributes(rest);
+  const validAttrProps = Object.assign(
+    _.pick(rest, reactRouterLinkProps),
```
alex-wilmer: rather than combine `validAttrProps with "valid router" props, can we spread the valid router props separately?

cheapsteak: done

alex-wilmer:
awesome thanks

PR #1747, opened 2017-06-02
✨ PRTL-1232 Do not show "Add filtered files to cart" when no filters are present

cheapsteak:


alex-wilmer:
If I'm reading the ticket correctly it's asking for something slightly more specific--the existence of a file filter, not the existence of any filter

https://jira.opensciencedatacloud.org/secure/attachment/14730/1.2-Prod.png

alex-wilmer:
I don't remember why it was done that way and if it's still makes sense to do it like that. @phuongmy

Jephuff:
maybe it would be a better to just check if filtered files is less then files?

PR #1750, opened 2017-06-02
✨ [PRTL-1263] remove 'most' when it's all the cases

Jephuff:


alex-wilmer:
not really related to this pr specifically, but I think the text should not mix "top" and "most" if they mean the same thing. suggestion:

> xx most mutated cases and xx most mutated genes

Jephuff:
Makes sense, I'll merge this and we can make a follow up change when we get the go ahead :)

PR #1754, opened 2017-06-05
♻️ [PRTL-1259] pass width instead of bandwidth

Jephuff:
Removes the bandwidth calculation in all the charts.
Also, cancer dist chart to show 20


alex-wilmer:
this looks really good! can you quickly take a look at the exploration page / gene tab and reduce the size of your window? Is the barchart shrinking and survival plot not? should we make a ticket for that?

Jephuff:
I think I found the problem, I can just pop it in here (it's only 2 lines)

PR #1756, opened 2017-06-05
✨ PRTL-1261 update copy for survival plot without data

cheapsteak:


cheapsteak:
Closing because I can't figure out how to fix the lint errors for just those lines
Cherry picking and recommitted here:
https://github.com/NCI-GDC/portal-ui/pull/1768

PR #1757, opened 2017-06-05
🐛 PRTL-1274 401/403 check if logged in first b4 popup

alex-wilmer:
- if the user state exists, then show session expired
 - also forceLogout the user

### DIFF
--- a/src/packages/@ncigdc/components/Header.js
+++ b/src/packages/@ncigdc/components/Header.js
@@ -19,6 +19,7 @@ import LoginButton from "@ncigdc/components/LoginButton";
 import UserDropdown from "@ncigdc/components/UserDropdown";
 import Hidden from "@ncigdc/components/Hidden";
 import { setModal } from "@ncigdc/dux/modal";
+import { forceLogout } from "@ncigdc/dux/auth";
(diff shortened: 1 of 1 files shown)


alex-wilmer:
this needs to be tested in nightly. we were showing Session Expired modal on 403.. are these things related? isn't 403 permission denied / forbidden which would never be caused by an expired session. @cy ?

cy:
copied it from submisson https://github.com/NCI-GDC/submission-ui/blob/next/src/js/app.js#L55 which I also wrote 😅 I think, we saw 403 come back from /auth there at some point and wanted to force logout for that in addition to 401. Difference is, we directly show the error's message https://github.com/NCI-GDC/submission-ui/blob/795c83b5ec1e71e7589ec44bedbd19f0536e939a/src/js/views/login.js#L195 instead of 'Session expired' for both statuses. Do the same here? Or just ingore 403 if that's never gonna happen.

alex-wilmer:
@cy I think this is good enough for the time being, or that is, it's better than having it trigger on page load. I removed the 403, which is a nuance we can handle later on. perhaps we can get this in soon as vincent mentioned

PR #1763, opened 2017-06-05
:bug: PRTL-1270 Fix CurrentFilters number display, facet reset for range

cy:
note it's this grey rest button not the clear all:
screen shot 2017-06-06 at 11 44 42 am


cy:
changed to using`removeFilter` in filters/index.js r? @Jephuff

PR #1765, opened 2017-06-06
🐛 [PRTL-1235] remove filter when removing facet

Jephuff:


src/packages/@ncigdc/utils/filters/index.js
```diff
@@ -232,4 +233,20 @@ export const makeFilter: TMakeFilter = (fields, returnString = true) => {
   return returnString ? JSON.stringify(filters) : filters;
 };
 
+type TRemoveFilter = (field: string, query: TUriQuery) => TUriQuery;
```
cy: I did the same thing here https://github.com/NCI-GDC/portal-ui/pull/1763/commits/189d6aff0a79992345a722d3b5883039f4413b3f#diff-766ae864db196552a908d837bfce2449R26 except wanted to detect inCurrent at the same time. Should I use yours? Hmmmm

Jephuff: hmm, I could make the removeFilter function return that information too? seems like it could be a common use case.

Jephuff: or you could use this + `inCurrentFilters`?

cy: I'll use this + inCurrentFilters 👍

PR #1766, opened 2017-06-06
🔧 prettier -> trailingComma: all, singleQuote: true

alex-wilmer:


alex-wilmer:
Just throwing this out there, but I think single quotes would be nice too. That's what we've had since as long as I've been here and I always type singles (except for when I always type backticks, but y'know)

cy:
+1 single quotes, y`know

shanewilson:
we can do this after the code freeze wednesday release thing

shanewilson:
also you guys should pick out all the changes you want now cause we really shouldn't do this sort of thing very often/ever again

alex-wilmer:
Those are the only two I care about.

cheapsteak:
+1000 (merge conflict aside)

alex-wilmer:
https://github.com/NCI-GDC/portal-ui/commit/afd19d94fa89ae17de0537b760ebe930cbd7a15f

PR #1769, opened 2017-06-06
🐛 [PRTL-1275] update annotations link in repo page

Jephuff:


alex-wilmer:
This looks good for case, but won't this not make sense for files? We don't have an appropriate filter for this yet since you can't filter annotations bye file

Jephuff:
hmm, good point, that must have been why the link was missing. I'll ask about this in slack 👍

Jephuff:
Marking this as ready for review again. I think it makes sense to merge this and then make another pull request for file if needed in the future.

PR #1772, opened 2017-06-06
🐛💄 PRTL-1087 Firefox: Fix OncoGrid fullscreen view

cheapsteak:


Jephuff:
Background is still black in my firefox :( 
image

cheapsteak:
Weird
It's showing up as white on mine (black at first then flickers to white)

Jephuff:
ahh, the one in explore worked for me, but the one on the project page didn't.

cheapsteak:
I see, it seems like the fullscreen styles weren't being applied when going full screen from the projects page...

moved the background color to the container instead (assumption being it'll always be in a container that has a white bg)

cheapsteak:
updated, could you take another look when the build'd done tomorrow?

PR #1777, opened 2017-06-07
✨🐛 PRTL-1281 keep state order in arrange dropdown

alex-wilmer:


Jephuff:
I don't think that the active columns should be sorted to the top, makes it a bit annoying to use ( at least for me). I think we should match prod, thoughts?

alex-wilmer:
i tried to match prod but couldn't figure it out. I'll give it another show

PR #1780, opened 2017-06-07
✨ PRTL-1070 warning + link + tooltip hidden ssm

alex-wilmer:


alex-wilmer:
screen shot 2017-06-07 at 2 07 39 pm

cheapsteak:
image

Is this behaving as expected or is it not updating?

alex-wilmer:
the behaviour is correct but maybe the language isn't perfect. the numerator shows all the mutations that are "in view" (within the zoomed in range). The warning icon appears when there are mutations that occur outside of the fully zoomed out range, meaning they will never be seen

shanewilson:
what does clicking the warning sign do?

cheapsteak:
The proximity of the warning icon to the "Viewing x/y Mutations" makes it seem like they are related
If the warning is related more closely to the transcript, perhaps it should be moved here instead?
image

alex-wilmer:
@shanewilson  it's a link. if there's one it takes you to entity, more takes you to exploration filtered on those ids

shanewilson:
Oh I'm not sure about that behaviour - I assumed it would expand the view to include it. Going to another page might be confusing?

alex-wilmer:
This wasn't specced out. was the result of a conversation I had with @junjun-zhang

alex-wilmer:
that's interesting though, It could do that.. up the `max` variable to get the mutation to show on the chart

shanewilson:
that can be a new feature if this is already done - or we can disable the warning until we spec out the behaviour.

alex-wilmer:
Based on feedback from @junjun-zhang 
screen shot 2017-06-07 at 4 32 30 pm

PR #1786, opened 2017-06-08
✏️ 1243 update warning wording

cy:
didn't catch that one sentence changed 😨

### DIFF
--- a/src/packages/@ncigdc/components/Modals/FirstTimeModal.js
+++ b/src/packages/@ncigdc/components/Modals/FirstTimeModal.js
@@ -31,7 +31,9 @@ const FirstTimeModal = () =>
     </p>
     <p>
       <b>WARNING</b>
-      : Please read the
+      : Data in the GDC is considered provisional as the GDC applies
+      state-of-the art analysis pipelines which evolve over time. Please read
+      the
       {' '}
       <ExternalLink
         hasExternalIcon={false}


src/packages/@ncigdc/components/Modals/FirstTimeModal.js
```diff
@@ -20,17 +20,18 @@ const FirstTimeModal = () => (
...
+
       <ExternalLink
         hasExternalIcon={false}
         href="https://docs.gdc.cancer.gov/Data/Release_Notes/Data_Release_Notes"
       >
         GDC Data Release Notes
-      </ExternalLink>
-      {" "}
-      prior to accessing this web site as the Release Notes provide details about data updates, known issues and workarounds.
+      </ExternalLink>&nbsp;
```
shanewilson: do we want a `&nbsp;` here?

alex-wilmer: both `{" "}` and `&nbsp;` make sense to me in this context

shanewilson: `&nbsp;` is different though isn't it? the line won't wrap around it. If it is just used to get around prettier I don't think we should use it.

alex-wilmer: @cy is that why you changed it? because prettier weirdness?

cy: yeah

cy: I didn't like the extra newline just for {" "} but I can change it back :woman_shrugging:

shanewilson: I think messing with prettier output sends us down a dangerous path of bike shedding.

src/packages/@ncigdc/components/Modals/FirstTimeModal.js
```diff
@@ -29,8 +29,9 @@ const FirstTimeModal = () => (
         GDC Data Release Notes
       </ExternalLink>
       {" "}
-      prior to accessing this web site as the Release Notes provide details about data updates, known issues and workarounds.
+      prior to accessing this web site as the Release Notes provide details about data updates, and known issues and workarounds.
```
shanewilson: I don't think you need this comma since there are only two things?

alex-wilmer: oh this came back? I remember asking you to remove that `and`.. did someone request this to come back?

shanewilson: or if `known issues and workarounds` are two things I don't think you need the extra `and`

alex-wilmer: I think it should read... "details about data updates, known issues and workarounds"

alex-wilmer: left it is there

<-----------

cy: where is this arrow pointing

cy: oh I see the extra 'and', didn't read it carefully, copy and pasted it from the ticket, which was a huge text blob. From the diff, looks like the change is adding this sentence in the warning `Data in the GDC is considered provisional as the GDC applies state-of-the art analysis pipelines which evolve over time.`

src/packages/@ncigdc/components/Modals/FirstTimeModal.js
```diff
@@ -40,8 +40,7 @@ const FirstTimeModal = () =>
         GDC Data Release Notes
       </ExternalLink>
       {' '}
-      prior to accessing this web site as the Release Notes provide details
-      about data updates, known issues and workarounds.
+      prior to accessing this web site as the Release Notes provide details about data updates, and known issues and workarounds.
```
shanewilson: ~and~

src/packages/@ncigdc/components/Modals/FirstTimeModal.js
alex-wilmer: oh haha the side by side view.. i'm pointing to the way it was before. Did someone request to add this "and" back? I don't think it reads well

src/packages/@ncigdc/components/Modals/FirstTimeModal.js
```diff
@@ -31,7 +31,9 @@ const FirstTimeModal = () =>
     </p>
     <p>
       <b>WARNING</b>
-      : Please read the
+      : Data in the GDC is considered provisional as the GDC applies
+      state-of-the art analysis pipelines which evolve over time. Please read
+      the
```
cy: the on one line, thanks prettier 🚲

PR #1788, opened 2017-06-08
🐛 guard on null summary.file_count case index

alex-wilmer:


cy:
but it crashes the whole page, showing 0 still shows the data issue. Maybe show 'null'?

shanewilson:
I don't want to merge this because I'm concern the issue will just be ignored with people thinking it was just a UI issue - when I get confirmation it is being worked on I guess we can merge it

alex-wilmer:
sure you want to merge this?

shanewilson:
I don't think it matters - either way - should never be an issue but I don't think it causes any problems.

PR #1789, opened 2017-06-08
🐛 [PRTL-1296] use auth url

Jephuff:


src/packages/@ncigdc/components/Project.js
```diff
@@ -110,7 +110,7 @@ const Project = (
           <DownloadButton
             disabled={!biospecimenCount}
             filename={`biospecimen.project-${projectId}`}
-            url={urlJoin(process.env.REACT_APP_API, 'cases')}
+            url={urlJoin(process.env.REACT_APP_GDC_AUTH, 'api/cases')}
```
shanewilson: add a new ENV called `REACT_APP_GDC_AUTH_API` since 99% of the time it needs the /api part - the only time it doesn't is for the `/user` request

Jephuff: would `REACT_APP_GDC_AUTH_API` ever diverge from `REACT_APP_GDC_AUTH`? maybe it would be worth adding a constants file and just adding `/api` there?

shanewilson: `REACT_APP_GDC_AUTH_API` is `REACT_APP_GDC_AUTH` + `/api`

Jephuff: moved them to the constants file

PR #1794, opened 2017-06-08
🐛 case entity, ssmstable filter by case id

alex-wilmer:


shanewilson:
@alex-wilmer does this need to be confirmed by anyone else?

alex-wilmer:
I believe so. It's too big of a change, someone will notice and they'll think its a bug

alex-wilmer:
i'll just close this for now. something to keep in mind for later

PR #1797, opened 2017-06-09
✨ [PRTL-1275] add link for 1 annotation

Jephuff:
Not sure if we want to merge this, but phuong-my requested it.
Adds link if there is 1 annotation in the file repo page.


alex-wilmer:
This can be fixed with sets now right?

Jephuff:
yep, it will need a little api work too though.

PR #1800, opened 2017-06-09
💄 flexwrap pie charts

alex-wilmer:


alex-wilmer:
image

Jephuff:
Might be better to left align, this looks a bit strange
image

File tab will need an update as well, it probably should hide pie charts when there are spaces available (probably also needs 'space-between')
image

alex-wilmer:
I don't know, I think its worth getting in now. Simple fix for:

image

@Jephuff not sure I feel that left aligned is better than centered in this case

shanewilson:
the `show more` thing in jeffery's screenshot doesn't really make sense anymore.

alex-wilmer:
doesn't it? that adds more pies.

alex-wilmer:
in any case, this pr doesn't impact that

shanewilson:
why don't the pies just fill in the empty space? It looks weird

Jephuff:
Maybe instead of wrapping, extra pies should be hidden? based on the file tab it seems the motivation was to keep it 1 row.

alex-wilmer:
I don't care I just want it to not break

shanewilson:
hiding extra pies is fine but falls under features vs bug fix

PR #1801, opened 2017-06-09
💄 PRTL-1303 survival column touchups

alex-wilmer:


alex-wilmer:
Is it not supposed to be?

image

shanewilson:
it is

shanewilson:
I'm assuming this mean survival is right aligned since it wasn't in the screenshot

alex-wilmer:
yup! I just grabbed a random screen from prod

PR:

image

PR #1804, opened 2017-06-09
Remove eslint-disable lines

Jephuff:


src/packages/@ncigdc/components/FacetSelection.js
```diff
@@ -1,118 +1,110 @@
 /* @flow */
-/* eslint-disable */
 
-import React from "react";
-import _ from "lodash";
-import { css } from "glamor";
+import React from 'react';
+import _ from 'lodash';
+import { css } from 'glamor';
```
shanewilson: shouldn't these be `"`?

shanewilson: oh wait we went the other way - so was this file missed? I don't get how files are missed when running `eslint . --fix`

shanewilson: oh again - I am dumb I get it

Jephuff: 👍

PR #1816, opened 2017-06-12
♻️🐎 only run explore case query when rendered

alex-wilmer:


alex-wilmer:
This makes the mutation / gene tabs in exploration significantly faster on the first load

alex-wilmer:
Yeah I was thinking of doing that, but was hesitant in case we were on the verge of going fully relay modern. 

agreed, though

PR #1817, opened 2017-06-12
🛠 only commit staged changes

Jephuff:
Proposed change to precommit: This stashes unstaged changes and then restores them so you can do partial commits.


shanewilson:
hmm - have you read through these?

https://github.com/okonet/lint-staged/issues/62
https://github.com/facebookincubator/create-react-app/pull/1759

they talk about merge conflicts coming about after the pop.

Can you setup sourcetree to use --no-verify?

Jephuff:
I think the merge conflicts from popping would be pretty rare, and only when you are committing a file partially.
But If i'm the only one who has a workflow like this, I can just just skip the hook locally :)

shanewilson:
well we could also just remove the hook. if everyone has auto formatting on their editors then it does nothing

cheapsteak:
Partial commits would also be useful for me as well

Just set up format on save

Jephuff:
Is anyone relying on the precommit to lint their files? we could leave `eslint` in so it blocks unlinted commits.

Jephuff:
@cheapsteak no complaints from me. We could also gitignore the settings file now that we don't need to support node_modules?

cheapsteak:
either way works for me, whatever would allow me to stop having to stash and unstash my settings between branches and pulls

shanewilson:
I would prefer we go with simple and not in anyones way. It looks like the easiest solution is to remove lint-staged. I'm fine with this since I get no benefit from it anyway - it was mostly because I thought @Jephuff didn't want to use auto-format on his editor.

PR #1821, opened 2017-06-13
SV-697 update links to use cancer.gov

Fgerthoffert:
gdc.nci.nih.gov -> gdc.cancer.gov
gdc-portal.nci.nih.gov -> portal.gdc.cancer.gov
gdc-docs.nci.nih.gov -> docs.gdc.cancer.gov

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "name": "OICR",
     "url": "https://github.com/NCI-GDC/"
   },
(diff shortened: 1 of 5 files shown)


Fgerthoffert:
No, it's not critical since redirection will still work.
Sorry for the wrong commit name, didn't update from default github :(

alex-wilmer:
np on commit names. we can squash it down and add the approporiate emoji 😄

PR #1823, opened 2017-06-13
♻️ fully modern SsmsTable

alex-wilmer:
definitely a WIP still but the data is loading correctly! some things like `relay.variables` and `onReadyStateChange` no longer exist, so will have to figure out how to put those back.

also the environment object may need to be tweaked to match the custom layer we're injecting.


alex-wilmer:
This should be good to go now. Good first step to moving things over further along to modern

PR #1824, opened 2017-06-13
🐛 fix dupe key for banner notification

alex-wilmer:
fixes this:

image

### DIFF
--- a/src/packages/@ncigdc/dux/bannerNotification.js
+++ b/src/packages/@ncigdc/dux/bannerNotification.js
@@ -33,7 +33,7 @@ let initialState = [
   {
     components: ['PORTAL'],
     level: 'INFO',
-    id: 0,
+    id: 'initial_banner',
     dismissible: true,
     message: <DismissibleBanner />,
   },


src/packages/@ncigdc/dux/bannerNotification.js
```diff
@@ -33,7 +33,7 @@ let initialState = [
   {
     components: ['PORTAL'],
     level: 'INFO',
-    id: 0,
+    id: 'initial_banner',
     dismissible: true,
     message: <DismissibleBanner />,
```
cy: might need `dismissed: false` so the `<DismissibleBanner />` still shows? not sure

alex-wilmer: I don't think we do.. but I like more verbose initial state if that's an important thing

shanewilson: what is this initial state for?

shanewilson:
should this be into master? the jira ticket says critical

alex-wilmer:
sure

src/packages/@ncigdc/dux/bannerNotification.js
cy: it's for setting that `Can't find your data?` dismissible notification that's hardcoded and not coming from the endpoint

alex-wilmer:
Yeah it was just a key error. the nightly banner also had `id : 0`

PR #1825, opened 2017-06-14
Isolated component route

alex-wilmer:
I've wanted to do this for a while.. a quick way to test a component by itself.

this makes a route at `/components/:component_name` and loads just that component. Should be a really good way to test default props and performance per component.. lots of things really. I was thinking about storybook for a while but this seems easier for now

image


shanewilson:
what about story book?

shanewilson:
this looks like it is splitting component and query again

alex-wilmer:
This doesn't preclude storybook, which I have running locally, but that will require a lot more work. there's a number of issues with it and our app out of the box.

PR #1828, opened 2017-06-14
🐛 PRTL-838 tooltip positioning off in IE

cheapsteak:
moveTooltip needs to wait to be called after tooltip's offsetHeight has the correct value

### DIFF
--- a/src/packages/@ncigdc/uikit/Tooltip/GlobalTooltip.js
+++ b/src/packages/@ncigdc/uikit/Tooltip/GlobalTooltip.js
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import './style.css';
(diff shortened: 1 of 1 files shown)


src/packages/@ncigdc/uikit/Tooltip/GlobalTooltip.js
```diff
@@ -11,12 +12,16 @@ class GlobalTooltip extends Component {
...
-  moveTooltip = event => {
-    this.globalTooltip.style.left = `${event.pageX}px`;
-    this.globalTooltip.style.top = `${event.pageY -
-      this.globalTooltip.offsetHeight -
-      15}px`;
-  };
+  moveTooltip = _.throttle(event => {
+    const left = event.pageX;
+    const top = event.pageY;
+    setTimeout(() => {
```
Jephuff: Does this fix it, or just make the edge case edgier? Would it make sense to add a componentDidUpdate and add/remove the listener based on `props.tooltip.Component`?

cheapsteak: This either fixes it or makes it edgier to the point of being undifferentiable from being fixed :p

I'm not sure if the position would be correct with the current global tooltip solution if the listener is added after componentDidUpdate. The mouse move event may happen before that lifecycle hook is called

Jephuff: yeah, you are right.
Wanna switch this from using top/left to using transforms well your in here? :D

cheapsteak: Worried about potential bugs  
This tooltip has position absolute so it wouldn't cause any DOM reflows  
Would recommend we prioritize minimizing bugs unless this proves to be a performance bottleneck

Jephuff: Fair. Realized we also don't need the height to calculate tooltip position, but that would be a bit more of a rewrite, so maybe in the future we can do that :)

PR #1836, opened 2017-06-15
🐛♻️ PRTL-1234 survival svg export

cheapsteak:


src/packages/@ncigdc/utils/wrapSvg.js
```diff
@@ -171,7 +171,10 @@ export const wrapSvg: TWrapSvg = ({
     margins.top,
     beforeObject.height,
   ])})">
-        ${svg.innerHTML}
+        ${svg.innerHTML.replace(
+          /url\((?:['|"]*)(?:http|https):\/\/(?:.+)(#.+)(?:['|"]*)\)/g,
```
Jephuff: that regex seems a little complicated, any reason it can't just be `/url\(['"]?https?:\/\/[^#]+(#.+)['"]?\)/g`?

cheapsteak: haven't tested that one but looks like it should work, I'll update it

PR #1838, opened 2017-06-15
🐛 PRTL-1327 resetting facets should not reset current tab

cheapsteak:


src/packages/@ncigdc/components/Aggregations/FacetResetButton.js
```diff
@@ -26,26 +28,31 @@ const StyledLink = styled(Link, {
...
+  };
   const inCurrent = fieldInCurrentFilters({
     currentFilters: currentFilters.content || [],
     field,
   });
   return (
     <StyledLink
       style={{ display: inCurrent ? 'inline' : 'none', ...style }}
-      query={inCurrent ? newQuery : {}}
+      query={newQuery}
```
shanewilson: shouldn't need to import router to do this - can't this be handled by the link?

cheapsteak: This doesn't seem to work (nothing happens when you click the reset button)

```js
const FacetResetButton = ({
  field,
  currentFilters,
  style,
  ...props
}) => {
  const newFilters = removeFilter(field, currentFilters);
  const newQuery = {
    offset: 0,
    filters: newFilters,
  };
  const inCurrent = fieldInCurrentFilters({
    currentFilters: currentFilters.content || [],
    field,
  });
  return (
    <StyledLink
      merge="replace"
      style={{ display: inCurrent ? 'inline' : 'none', ...style }}
      query={newQuery}
    >
      <ShadowedUndoIcon /><Hidden>reset</Hidden>
    </StyledLink>
  );
};
```

Jephuff: I think you either need to use `toggle` and pass a filter that has just the values you want to remove, or maybe add a new merge function like `replaceAll` that would just be `q => q`

cheapsteak: Can we just use this? Works and can understand what's going on without having to know the intricacies of how `StyledLink` works

Jephuff: I have no objections.

PR #1843, opened 2017-06-16
Add analysis service integrations. Track Oncogrid & Survivalplot performance

cheapsteak:


src/packages/@ncigdc/utils/analytics/performanceTracker.js
```diff
@@ -0,0 +1,34 @@
+const startTimes = {};
+export const timer = {
+  time: label => (startTimes[label] = performance.now()),
+  timeEnd: label => {
+    if (!startTimes[label]) {
+      console.warn(`No start time was found for "${label}"`);
+      return;
+    }
+    const duration = performance.now() - startTimes[label];
```
Jephuff: should probably remove startTimes here, to prevent calling timeEnd multiple times.

src/packages/@ncigdc/utils/analytics/performanceTracker.js
```diff
@@ -0,0 +1,34 @@
+const startTimes = {};
+export const timer = {
+  time: label => (startTimes[label] = performance.now()),
```
Jephuff: If you go with my suggestion for line 9, It might also be good to at least warn if startTimes[label] already exists.

src/packages/@ncigdc/utils/analytics/performanceTracker.js
cheapsteak: done

PR #1847, opened 2017-06-20
🐛 PRTL-1337 use fetched transcripts if none provided

alex-wilmer:


src/packages/@ncigdc/containers/Lolliplot.js
```diff
@@ -773,7 +777,7 @@ export const LolliplotQuery = {
             symbol
             canonical_transcript_id
             transcripts {
-              hits(first: 1 filters: $isCanonical) {
+              hits(first: 99) {
```
shanewilson: what does this fix?

alex-wilmer: transcript dropdown above the viewer

shanewilson: what would be the real solution to this? another api call?

alex-wilmer: a paginated dropdown?

PR #1848, opened 2017-06-20
🐛 tooltip throttle gives wrong pos + choppy feel

alex-wilmer:


cheapsteak:
please do test this in IE for https://jira.opensciencedatacloud.org/browse/PRTL-838
image

(only seems to affect IE)

cheapsteak:
The IE bug was fixed by the previous PR that added a setTimeout
Unsure if this PR causes regression

alex-wilmer:
I think this is very important, but the current state has gone too far. I'll just reduce the throttle to 16

PR #1852, opened 2017-06-21
🐛 [PRTL-1344] filter oncogrid by impact exists

Jephuff:


Jephuff:
I think @junjun-zhang tested it

PR #1853, opened 2017-06-21
🐛 PRTL-1307 Don't add "null" to filters when clicking bar

cheapsteak:


src/packages/@ncigdc/components/Explore/GenesTab.js
```diff
@@ -77,10 +78,13 @@ export default compose(
       );
       push({
         pathname: '/exploration',
-        query: {
-          ...query,
-          filters: JSURL.stringify(newFilters),
-        },
+        query: _.omitBy(
```
Jephuff: Maybe use `removeEmptyKeys` from utils/uri. It removes more then just undefined and null;

cheapsteak: Good idea, updated

PR #1858, opened 2017-06-23
🐛 [PRTL-1360] filter pie aggregations with all filters

Jephuff:


src/packages/@ncigdc/containers/RepositoryPage.js
```diff
@@ -265,6 +267,8 @@ export const RepositoryPageQuery = {
           cases {
             aggregations(filters: $filters aggregations_filter_themselves: false) {
               ${CaseAggregations.getFragment('aggregations')}
+            }
+            pieAggregations: aggregations(filters: $filters aggregations_filter_themselves: true) {
```
shanewilson: this is how it is done on the exploration page? just want to make sure they are consistent

Jephuff: yep, but they are called `pies` there, I'll update to match.

shanewilson: there was also a bug where the label would say 1 files0 or 1cases0 if there was only 1 item in the pie

shanewilson: probably a typo

shanewilson:
there was also a bug where the label would say 1 files0 or 1cases0 if there was only 1 item in the pie
probably a typo

Jephuff:
yep! found it, it was `0` when it should have been `''`'. Is there a ticket for that?

shanewilson:
I don't think so but it was found at the same time this bug was so I guess they can be fixed together

PR #1863, opened 2017-06-23
🔥 PRTL-1358 remove ssmsBarChart

alex-wilmer:


alex-wilmer:
Not a usual removal case. I think it's worth keeping the code here, but don't render the component

shanewilson:
so this is just removing the chart and then there is another ticket for adding the spark line?

alex-wilmer:
that's my plan. should those go together?

alex-wilmer:
I don't think there was a ticket for sparkline

shanewilson:
no that's fine - just being clear

shanewilson:
1352 has the screenshot you made

PR #1866, opened 2017-06-23
✨ SparkMeter in ssmsTable

alex-wilmer:


shanewilson:
does this need to be added to more tables? did we have that bar chart on any entity pages?

alex-wilmer:
it could be added to the genes table, but we're keeping that bar chart afaik

shanewilson:
just the bar charts that have mutations id on the x axis need to be removed

alex-wilmer:
right, so this good then, yes?

PR #1879, opened 2017-06-27
🔧 Don't load newrelic in "development"

cheapsteak:


public/index.html
```diff
@@ -24,7 +24,12 @@
   <title>%REACT_APP_WEBSITE_NAME%</title>
   <link href="//maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" integrity="sha384-wvfXpqpZZVQGK6TAh5PVlGOfQNHSoD2xbE+QkPxCAFlNEevoEH3Sl0sibVcOQVnN"
     crossorigin="anonymous">
-  <script src="/newrelic.js"></script>
+    <script>
+      "%NODE_ENV%" !== "development" && (function () {
```
shanewilson: `"%NODE_ENV%" === "production"` ?

cheapsteak: 👍

Jephuff:
how would you feel about ```<script>
if( "%NODE_ENV%" === "production") {
  var script = document.createElement('script');
  script.src = '/newrelic.js';
  document.body.appendChild(script);
}
</script>
```?
or does it need to be loaded before the site loads? 
Just confirmed about this 
![image](https://user-images.githubusercontent.com/3953093/27606273-a3e1fa20-5b4d-11e7-8344-c1a413ac5d81.png)
Future us might get confused whats happening there

This is what it showed before
![image](https://user-images.githubusercontent.com/3953093/27606305-c0b972e0-5b4d-11e7-8d35-7ddb6ea4c05f.png)

**Comment by cheapsteak** [comment]
I think one of the things it tries to measure is how fast our bundles load
This would negate that

**Comment by Jephuff** [comment]
ahh ok, how about moving that conditional into the script than? could just add something like `<script>NR = "%NODE_ENV%" !== "production"</script>` above it?

**Comment by cheapsteak** [comment]
Not sure what you mean?

**Comment by Jephuff** [comment]
put ```
<script>NR = "%NODE_ENV%" !== "production"</script>
<script src="/newrelic.js"></script>
```
in the index.html file and then in the newrelic.js script you can do the conditional like `NR && ...`

cheapsteak:
Ohhh, interesting idea

PR #1898, opened 2017-07-05
[minor] :pencil: PRTL-1389 Update pie titles to singular

cy:


Jephuff:
`Primary Sites` in explore should probably be `Primary Site` right?

cy:
@Jephuff missed that, updated

PR #1909, opened 2017-07-07
🐛 PRTL-1398 1 row per transcript, fix strand tsv export value, table name

cheapsteak:


src/packages/@ncigdc/containers/ConsequencesTable.js
```diff
@@ -131,6 +136,7 @@ const ConsequencesTableComponent = compose(
       style={{ width: '100%', minWidth: 450 }}
       data={dataRows}
       prefix={paginationPrefix}
+      entityName={'entries'}
```
cy: looks like the ticket wants 'consequences'

cheapsteak: There are only 4 different consequences on the table though.. I think it's better to keep it ambiguous. Will ask PM

shanewilson: because it is a table of transcripts

cheapsteak: another way to look at it is that it's a table of entries that happen to have unique transcripts, non-unique consequence types and aa variants

shanewilson: everything is a table of entries - it is like calling them widgets or items.

cheapsteak: I think the purpose of the the table is to convey "consequences", which is a combination of all the data of its columns. Transcripts also serve to convey consequences. Changing the way rows are aggregated such that transcripts are the only values unique per row doesn't change the purpose of the table

That said, I _am_ also open to renaming this to `Showing 1 - 10 of 27 items`

cy: what about just not naming it `Showing 1 - 10 of 27`

cheapsteak: That would also work

I just noticed though that we do refer to these as entires in the pagination size selection, which is used throughout the portal 
image

src/packages/@ncigdc/containers/ConsequencesTable.js
```diff
@@ -77,7 +77,12 @@ const ConsequencesTableComponent = compose(
           consequence: transcript.consequence_type,
           coding_dna_change: transcript.annotation.hgvsc,
           strand: transcript.gene.gene_strand
-            ? strandIconMap[transcript.gene.gene_strand.toString(10)]
+            ? <span>
+                {strandIconMap[transcript.gene.gene_strand.toString(10)]}
+                <ForTsvExport>
```
cheapsteak: That seems like it would require more work to support column sorting and column selection.

Would also be made obsolete with the TSV export endpoint that's in the backlog

Also not a fan of "if you're going to be making [small change], then you might as well make [larger change]". An easy way to have a PR snowball out of scope

src/packages/@ncigdc/containers/ConsequencesTable.js
cy: yeah, feel like `entries` isn't needed there either. @phuongmy?

src/packages/@ncigdc/containers/ConsequencesTable.js
Jephuff: I guess I mean more if your are going to do more work to support the temporary solution, why not do the easy solution that can be permanent :) But I won't block this PR about it or anything, so it's up to you.

src/packages/@ncigdc/containers/ConsequencesTable.js
phuongmy: I would keep "Show 10 entries" as it's very standard in all our tables and display `Showing 1 - 10 of 27` as suggested to avoid the debate about consequences/transcripts. Thanks!

src/packages/@ncigdc/containers/ConsequencesTable.js
cheapsteak: I think it will potentially cause confusion to have some TSV export buttons export all data, and other TSV export buttons only export the current page, with no way for the user to distinguish how the same button will behave, and no good way to communicate why the limitation exists for some tables and not others (users shouldn't have to know about client-side vs server-side pagination). 

Imo it's better to have predictable behavior, and defer this feature until we have the backend solution for TSV exports that will support exporting all rows for all tables

cheapsteak:
ready for re-review
image

image

PR #1910, opened 2017-07-07
✏️ PRTL-1396 clarify distinctness of mutations in tooltip

cheapsteak:


shanewilson:
yea why do we have that?

cheapsteak:
Feedback from Himanso (not sure who he/she is)

PR #1912, opened 2017-07-07
✏️PRTL-1403 Gene Page - Re-add the gene symbol on the consequence column

cheapsteak:


phuongmy:
Could we disable the Gene link when we are already in the Gene entity page? I tried to click on it and it felt like nothing is happening... because I am already in the Gene entity page. 
I would suggest 2 options:
1- Disable gene link when user is already in the gene entity page
2- If the link is active and user clicks on it, it should scroll up to top page.
screen shot 2017-07-10 at 5 34 46 am

cheapsteak:
@phuongmy done, currently building on travis but this is what it will look like:
image

PR #1900, opened 2017-07-05
Prtl 1385 upload gene set

Jephuff:
https://github.com/NCI-GDC/portal-ui/pull/1914 includes a fix for the tooltip to keep it on the screen to prevent this
image

~Based on https://github.com/NCI-GDC/portal-ui/pull/1846 which requires https://github.com/NCI-GDC/gdcapi/pull/806~

PR #1914, opened 2017-07-07
✨ update tooltip positioning

Jephuff:
keeps the tool tip on screen.

Example(from https://github.com/NCI-GDC/portal-ui/pull/1900): 
image
image


shanewilson:
we don't support tabs?

Jephuff:
We do, I'll add that. It's actually in this PR https://github.com/NCI-GDC/portal-ui/pull/1900

PR #1918, opened 2017-07-11
🔧 trace hoc

alex-wilmer:
image

### DIFF
--- a/src/packages/@ncigdc/utils/trace.js
+++ b/src/packages/@ncigdc/utils/trace.js
@@ -1,0 +1,19 @@
+/* eslint-disable no-sequences */
+
+import React from 'react';
+
+export default (additionalMessage = '') => Component => props => (
+  additionalMessage && console.log(additionalMessage),
+  console.info(
+    `%ctracing ${Component.displayName} props:\n${'tracing '.replace(
+      /./g,
+      '=',
+    )}${Component.displayName.replace(/./g, '=')}=${'props:'.replace(
+      /./g,
+      '=',
+    )}`,
+    'color: rgb(22, 138, 96);font-weight: bold;',
+  ),
+  console.log(props),
+  <Component {...props} />
+);
--- a/src/packages/@ncigdc/utils/trace.js
+++ b/src/packages/@ncigdc/utils/trace.js
@@ -2,6 +2,28 @@
 
 import React from 'react';
 
+/**
+ *  Utility HOC designed to easily log received props.
+*/
+
+/* Recommended to be put on the global namespace in development mode:
+
+global.trace = process.env.NODE_ENV === 'production'
+  ? C => p => <C {...p} />
+  : require(<path_to_trace>).default;
+
+*/
+
+/* usage:
+
+  global.trace('Trying to fix issue #12')(Button)
+
+  compose(
+    global.trace() // string argument is optional.
+  )
+
+*/
+
 export default (additionalMessage = '') => Component => props => (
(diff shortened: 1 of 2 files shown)


src/packages/@ncigdc/utils/trace.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export default C => p => (
+  console.info(
+    `%ctracing ${C.displayName} props:\n==========================`,
+    'color: rgb(22, 138, 96);font-weight: bold;',
+  ), // eslint-disable-line
+  console.log(p), // eslint-disable-line
+  <C {...p} />
```
cheapsteak: How do you feel about this?

```
C(p)
```

Also let's avoid single character variable names eh? Particularly if they abbreviate to CP

alex-wilmer: `C(p)` will fail if it's a class. that's the one use case where JSX aka `React.createElement` is better. I'll make the names better though

alex-wilmer:
@cheapsteak do you know if it's possible to make this global? would be really nice to not have to add and remove imports for this. `global.trace` is that a thing?

cheapsteak:
Add this to index.js?

```js
global.trace = process.env.NODE_ENV === 'production' ? x => x : require('@ncigdc/utils/trace.js')
```

alex-wilmer:
hmm not working for me. I tried both global and window but still getting undefined.

cheapsteak:
It's passing travis tho?

alex-wilmer:
this just puts it on global. i rebased this into my other branch and `global.trace` / `window.trace` is undefined inside a component

cheapsteak:
Oh I think this definition needs to happen before importing `Root`

alex-wilmer:
@cheapsteak  woot this works!

```
compose(  
  global.trace
  soManyThings(),
  global.trace
)
```

maybe it should take an argument for extra logging info?

cheapsteak:
Yeah, would recommend to always require invocation to keep the api consistent

alex-wilmer:
@cheapsteak okay. string arg added + commented the util

cheapsteak:
What do you think about moving the branching logic into trace?

```js
(additionalMessage = '') => branch(
  process.env.NODE_ENV === 'production',
  renderComponent(Component),
  Component => props => (/*...*/)
)

global.trace = require('@ncigdc/utils/trace');
```
?

alex-wilmer:
ah yes that's clearly better

src/packages/@ncigdc/utils/trace.js
```diff
@@ -0,0 +1,38 @@
...
+  )
+*/
+
+export default (additionalMessage = '') => Component =>
+  branch(
+    () => process.env.NODE_ENV === 'production',
+    renderComponent(Component),
+  )(
+    props => (
+      additionalMessage && console.log(additionalMessage),
```
Jephuff: Could this all be in the console.info? 
something like 
```
console.info(
    `${additionalMessage ? `${additionalMessage}\n` : ''}%ctracing ${Component .displayName} props:\n${'tracing '.replace(
      /./g,
      '=',
    )}${Component .displayName.replace(/./g, '=')}=${'props:'.replace(
      /./g,
      '=',
    )}\n`,
    'color: rgb(22, 138, 96);font-weight: bold;',
    props
  )
```
It would make it easier to filter the logs.

alex-wilmer: sure good idea

PR #1920, opened 2017-07-11
1386 Hide UUIDs

cy:


src/packages/@ncigdc/modern_components/SsmsTable/SsmsTable.js
```diff
@@ -247,14 +251,9 @@ export default compose(
...
             mutation_uuid: (
-              <span>
-                <MutationLink uuid={x.ssm_id}>
-                  {x.ssm_id.substr(0, 8)}
-                </MutationLink>
-                <ForTsvExport>
-                  {x.ssm_id}
-                </ForTsvExport>
-              </span>
+              <ForTsvExport>
```
alex-wilmer: why just `ForTsvExport` here?

cy: the column is only for tsv export, the Th is `display: none`, could do `display: none` on span but went with `ForTsvExport` so it's clearly there only for export.

PR #1928, opened 2017-07-12
🔥 remove jquery

Jephuff:


src/packages/@ncigdc/utils/download/index.js
```diff
@@ -251,37 +251,26 @@ const download = ({ url, params, method = 'GET', altMessage = false }) => {
...
+  iFrame.__frame__loaded = false;
+
+  const form = document.createElement('form');
+  form.method = method.toUpperCase();
+  form.action = url;
+  form.innerHTML = fields;
+
+  getBody(iFrame).appendChild(form);
+
+  form.submit();
```
alex-wilmer: this can be run before the form dom content has loaded?

Jephuff: Theres no funny business(loading async stuff) going on inside the form so waiting shouldn't be required.
Definitely give it a test though, everything worked for me.

alex-wilmer: yeah it worked, was just curious if you knew about this detail. the more I think about it, the more it makes sense that the form submitting action isn't relevant to it actually being on the dom.

PR #1949, opened 2017-07-18
:bug: CreateSaveSetBtn: dont die when currentFilters value is not a string

cy:
My bad, the create save set button reuses the set_id between the repo and explore page if current filters only contains set id. However the check fails when a range facet is added and the type of value is not a string but a number. Fix by casting toString().


cy:
@Jephuff added on https://github.com/NCI-GDC/portal-ui/pull/1949/commits/7c81ae2a93c2d13a4f602282eab89864a70c09e4#diff-fd2bdae4ce149b040671f3a2fbf58bffR282, think it's not needed on https://github.com/NCI-GDC/portal-ui/blob/7c81ae2a93c2d13a4f602282eab89864a70c09e4/src/packages/%40ncigdc/components/CurrentFilters.js#L87 because should be short-circuted

PR #1950, opened 2017-07-19
♻️ PRTL-1447 className instead of data-attribute

cheapsteak:
As discussed in the developer channel, digging selectors out from the html is challenging due to excessive noise from inlined styles and numerous wrapper DOM elements. 

Challenge - find the selector:

image

Now repeat the process to find selectors for the parents of the element you wish to target to ensure that you won't get the element with the same class/attr in a different section.

Chrome's "Copy Selector" may give a better starting point, even if it shouldn't be directly used for targeting an element due to brittleness.


cheapsteak:
@alex-wilmer would you mind giving this a look over?

PR #1966, opened 2017-07-24
🐛 correct genesymbol component in currentFilters

alex-wilmer:


Jephuff:
Could we remove `/components/GeneSymbol.js`? or is it required when used inside none relayModern components?

alex-wilmer:
yup I can remove it. just have to do the same fix to a few more files

PR #1967, opened 2017-07-24
:bug: PRTL-1481 ArrangeColumns restores default in itself

cy:
'Restore Default' button in ArrangeColumns restored columns in the table, but did not in the dropdown itself. This PR fixes that but there's still an issue with the last item not being able to be dragged past the hidden items if the rest of the list is the same. Thoughts @alex-wilmer?

(changes outside of src/packages/@ncigdc/components/ArrangeColumns.js are just adding types or a setDisplayName)

### DIFF
--- a/src/packages/@ncigdc/components/ArrangeColumns.js
+++ b/src/packages/@ncigdc/components/ArrangeColumns.js
@@ -1,7 +1,7 @@
 // @flow
 import React from 'react';
 import { connect } from 'react-redux';
-import { compose, withState, pure } from 'recompose';
+import { compose, withState, pure, withPropsOnChange } from 'recompose';
 import ArrangeIcon from 'react-icons/lib/fa/bars';
 
 import { Row } from '@ncigdc/uikit/Flex';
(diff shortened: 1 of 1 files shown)


Jephuff:
This is mostly fixed in next already, I think the only problem there is that you have to close and open the dropdown to get it to reset.

cy:
oh my bad thought this was for master didn't check next

cy:
ok updated with next. Fixes having to close the dropdown to get it to reset. But the issue of not being able to drag the last non-hidden item past the hidden items is still there. Reason being, the only ids being stored in the redux store are the non-hidden ids. In `withPropsOnChange`, I'm trying to detect if the order returned to default in the store, but none of the hidden columns are there so needed to filter those out.

Jephuff:
They are stored, just in a different key `order` so I think all you need to do is setColumns again when tableColumns changed.
```
withPropsOnChange(
    ['tableColumns'],
    ({ tableColumns, entityType, setState }) => {
      setState({
        columns: tableModels[entityType]
          .slice()
          .sort(
            (a, b) =>
              tableColumns.order.indexOf(a.id) -
              tableColumns.order.indexOf(b.id),
          ),
      });
    },
```

cy:
@Jephuff  that works! Updated 😊

PR #1968, opened 2017-07-25
:bug: PRTL-1449 hide oncogrid until render:all:end

cy:
works not sure if best approach

### DIFF
--- a/src/packages/@ncigdc/components/Oncogrid/OncogridWrapper.js
+++ b/src/packages/@ncigdc/components/Oncogrid/OncogridWrapper.js
@@ -52,12 +52,16 @@ function refreshGridState({
   setHeatMapMode,
   setShowGridLines,
   setCrosshairMode,
+  setIsLoading,
 }: {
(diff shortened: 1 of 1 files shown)


src/packages/@ncigdc/components/Oncogrid/OncogridWrapper.js
```diff
@@ -322,9 +322,11 @@ const OncoGridWrapper = compose(
         performanceTracker.begin('oncogrid:render');
         grid.render();
 
-        grid.on('render:all:end', () =>
-          performanceTracker.end('oncogrid:render', performanceContext),
-        );
+        grid.container[0][0].style.visibility = 'hidden';
+        grid.on('render:all:end', () => {
+          grid.container[0][0].style.visibility = 'visible';
```
Jephuff: `grid.container` should be a d3 object, so you could get the node with `grid.container.node()`

cy:
@Jephuff updated

PR #1973, opened 2017-07-27
PRTL-1456 use facets for custom facets and aggregations for preset

cy:
🔥 to generated relay parts

~~- fixing a problem with custom facets not loading when filters change sometimes~~ done but not super happy with it, kept reading from localstorage. will move custom facets to redux


cy:
@NCI-GDC/oicrdevs this is ready for review, simplifies custom facets to not use the non-standard generated relay queries and uses redux instead of directly read/write from localstorage. Future tickets could be: move to relay modern (back to generated queries!) and use graphql schema instead of /gql/_mapping (although the schema is huge).

alex-wilmer:
I was able to break it by selecting `demographic.days_to_death`, switching to the genes tab then switching back to cases

screen shot 2017-08-10 at 2 57 48 pm

cy:
ohhh didn't try a range facet. thanks

cy:
@alex-wilmer fixed

PR #1974, opened 2017-07-28
✨ [PRTL-1452] upload case set

Jephuff:


src/packages/@ncigdc/components/Modals/UploadSet/MappingTable/CaseMappingTable.js
```diff
@@ -0,0 +1,192 @@
...
+    const matchedCases = Object.entries(
+      matched.reduce((acc, hit) => {
+        const caseId = caseMap[hit].submitter_id.toUpperCase();
+        acc[caseId] = (acc[caseId] || []).concat(hit);
+        return acc;
+      }, {}),
+    )
+      .map(([caseId, submitted]) => {
+        const caseData = caseMap[caseId];
+        var temp = CASE_ID_FIELDS.map((idField, i) => {
```
alex-wilmer: I don't see this getting reassigned / used for hoisting?

src/packages/@ncigdc/utils/validateIds.js
```diff
@@ -1,30 +1,70 @@
-import { get } from 'lodash';
-
 import { fetchApiChunked } from '@ncigdc/utils/ajax';
 
+function _getIds(obj, path) {
+  const segment = path.shift();
```
alex-wilmer: [segment] = path ? kinda weird to see shift all of a sudden

src/packages/@ncigdc/components/Modals/UploadSet/MappingTable/CaseMappingTable.js
Jephuff: ahh yeah, that was a leftover from some debugging.

src/packages/@ncigdc/utils/validateIds.js
Jephuff: I was using it to also modify path, but I can refactor to not use shift 👍

PR #1975, opened 2017-07-28
♻️  use events for tooltips

Jephuff:
refactor for https://github.com/oncojs/oncogrid/pull/38

### DIFF
--- a/src/packages/@ncigdc/components/Oncogrid/mouseEvents.js
+++ b/src/packages/@ncigdc/components/Oncogrid/mouseEvents.js
@@ -75,7 +75,7 @@ export default function({ grid, setTooltip, trackLegends, push, dispatch }) {
     push(`/${type === 'donor' ? 'cases' : 'genes'}/${id}`);
   });
   grid.on('histogramClick', ({ domain: { id }, type }) => {
-    type = type === 'donor' ? 'case' : type;
+    const gdcType = type === 'donor' ? 'case' : type;
     push({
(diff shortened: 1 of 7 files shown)


src/packages/@ncigdc/components/Oncogrid/mouseEvents.js
```diff
@@ -0,0 +1,131 @@
...
+      <div dangerouslySetInnerHTML={{ __html: trackLegends[group] }} />,
+    );
+  });
+  grid.on('trackLegendMouseOut', () => setTooltip());
+
+  grid.on('trackClick', ({ domain: { id }, type }) => {
+    push(`/${type === 'donor' ? 'cases' : 'genes'}/${id}`);
+  });
+  grid.on('histogramClick', ({ domain: { id }, type }) => {
+    type = type === 'donor' ? 'case' : type;
```
cy: nitpick, use `const gdcType = type === 'donor' ? 'case' : type;` instead of overwriting type?

cy:
screen shot 2017-08-08 at 2 13 26 pm
possible to set a max width on the tips so it isn't so scary?

src/packages/@ncigdc/components/Oncogrid/mouseEvents.js
Jephuff: sure, I can do that :)

Jephuff:
Still pretty scary, but a little better :D 
image

PR #1976, opened 2017-07-28
⬆️ react-router-dom@4.1.2

alex-wilmer:
ok I got everything working *except* navigating away from the home page which for some reason throws:

```
createBrowserHistory.js:92 Uncaught TypeError: Cannot assign to read only property 'action' of object '#<Object>'
    at assign (<anonymous>)
    at setState (createBrowserHistory.js:92) 
```


alex-wilmer:
interesting.. when I add this to my branch where I removed the home classic route the error goes away. guessing it has something to do with loading the skeleton home page

https://github.com/NCI-GDC/portal-ui/pull/1964

alex-wilmer:
once that other branch is merged in can test this again

PR #1978, opened 2017-07-31
✨ Save and edit sets

Jephuff:
tickets: PRTL-1464, PRTL-1465, PRTL-1466, PRTL-1452, PRTL-1472, PRTL-1468, PRTL-1469, PRTL-1463, PRTL-1474

requires https://github.com/NCI-GDC/gdcapi/pull/831


alex-wilmer:
Apologies if some of these findings don't apply to your PR, just keeping track of them as I find them.

- when there's only one set check it automatically in modal for edit/remove
- uncheck all option? not just visible ones

api error when adding cases to a set where items had been removed:

```
{
  "data": {
    "sets": {
      "create": {
        "repository": {
          "case": null
        }
      }
    }
  },
  "errors": [
    {
      "locations": [
        {
          "column": 9,
          "line": 8
        }
      ],
      "message": "'NoneType' object has no attribute '__getitem__'"
    }
  ]
}
```

to reproduce: create a set with a few cases. uncheck everything, remove all cases from set, edit and try to add all the cases

- `Cannot convert undefined or null to object` when trying to download a TSV from 0 unmatched items in upload modal

```
src/packages/@ncigdc/utils/toTsvString.js:7

   4 | export default (xs: Array<Object>): string => {
   5 |   let result = '';
   6 |   if (xs) {
>  7 |     result = `${Object.keys(xs[0]).join('\t')}\n`;
   8 |     xs.forEach(x => {
   9 |       result += `${Object.keys(x).map(key => x[key]).join('\t')}\n`;
  10 |     });
```

- ability to name uploaded set instead of using 'input set' ?
- uploaded set not being saved in localstorage?

- I saved a match via TSV, then reuploaded it. it matched the same item but the resulting set had nothing found.

tsv file:

```
submittedCase ID	mappedCaseId	mappedProject
TCGA-AF-3912	TCGA-AF-3912	TCGA-READ
```

Jephuff:
@alex-wilmer I've addressed most of your feedback so far .
2 things, 
> ability to name uploaded set instead of using 'input set' ?
> uploaded set not being saved in localstorage?
I think this would be a spec change, right now they are ephemeral. Maybe the use case is uploading multiple tries without cluttering your save sets and then you can save the result with the `save/edit` button.

>I saved a match via TSV, then reuploaded it. it matched the same item but the resulting set had nothing found.
I'm wondering if this is a data issue, do you get results when enter the same case id in the suggestion facet?

cy:
checked for ya, that case TCGA-AF-3912 is in gdc_from_graph but not in case_centric

shanewilson:
Deploy preview processing.

Built with commit 2c9d0877957057c57eac52dba27c45540cb0b4a8

https://app.netlify.com/sites/rubber-tapper-lifts-72118/deploys/59974cac0752d027ccb0636c

src/packages/@ncigdc/dux/sets.js
```diff
@@ -0,0 +1,72 @@
...
+        ...action.payload.sets,
+      };
+    case sets.ADD_SET: {
+      const { label = '', id, type } = action.payload;
+
+      return {
+        ...state,
+        [type]: {
+          ...state[type],
+          [label]: id,
```
cy: can we switch this and have it keyed by id, actually? Working on the manage sets page, where the user can rename the set and would rather the key not change. Also the ids are guaranteed unique but not the labels.

Jephuff: yep, that makes sense! I'll update shortly

Jephuff: @cy done.

cy:
The 'Remove from existing set' is still expecting keyed by label? Because after using it on the 'Brain' set, got this:
screen shot 2017-08-22 at 10 48 44 am

src/packages/@ncigdc/dux/sets.js
Jephuff: @cy ok, should be done for real now

alex-wilmer:
Looks pretty good to me, tried to break stuff again and seems to be working

src/packages/@ncigdc/components/SetActions.js
```diff
@@ -0,0 +1,131 @@
...
+          </DropdownItem>
+          {hasSets &&
+            <DropdownItem
+              style={{ lineHeight: '1.5', cursor: 'pointer' }}
+              onClick={() => {
+                dispatch(
+                  setModal(
+                    <AppendSetModal
+                      field={field}
+                      title={`Add ${total} ${displayType} to existing set`}
```
alex-wilmer: `toLocaleString()` here?

PR #1980, opened 2017-08-02
🐛 1491, 1492, 1493, 1496 - annotations table / smart search fixes

alex-wilmer:


src/packages/@ncigdc/components/Header.js
```diff
@@ -36,6 +36,7 @@ const styles = {
 };
 
 const Header = compose(
+  withRouter,
```
Jephuff: This probably isn't needed if https://github.com/NCI-GDC/portal-ui/pull/1982 is merged right?

alex-wilmer: yea *shrug*

PR #1981, opened 2017-08-02
remove react-modal

Jephuff:


Jephuff:
https://github.com/NCI-GDC/portal-ui/pull/1982 fixes the problem so this isn't required.
Does anyone want to still remove `react-modal` or should I close this?

alex-wilmer:
I'm still in favor of removing it. just need to make sure we didn't lose any 508 compliance or something

PR #1987, opened 2017-08-03
💄 PRTL-1503 dnaChange for ssm entity header

alex-wilmer:


Jephuff:
min height is causing some issues
image

PR #1990, opened 2017-08-03
💄 PRTL-1505 cons table 'showing' in ssmrout

alex-wilmer:


src/packages/@ncigdc/modern_components/ConsequencesTable/ConsequencesTable.js
```diff
@@ -130,9 +130,11 @@ export default compose(
       canonicalTranscriptId,
       functionalImpactTranscript,
       theme,
+      context,
     }: TProps = {},
   ) =>
     <LocalPaginationTable
+      entityName={context}
```
Jephuff: Instead of passing this down, the default of `projects` should probably be removed from `LocalPaginationTable`. That was my bad, doesn't really make sense to have a default there :)

alex-wilmer: yeah I thought but that but it looked like maybe something was actually using that default

Jephuff: I think everything was updated. Might have been lost in the modern conversion?

alex-wilmer: ok i'll take a look and update

PR #1992, opened 2017-08-03
🐛 [PRTL-1509] handle undefined field in sort

Jephuff:


alex-wilmer:
commit says remove localCompare, but PR still uses it?

Jephuff:
@alex-wilmer oops, updated the PR title, but forgot to update the commit message. Updated now. I added it back because it wasn't the actual problem.

PR #2004, opened 2017-08-09
PRTL-1562 add 'distinct (unique)' to CancerDist mutation tooltip

cy:


alex-wilmer:
why do we need both distinct and unique?

alex-wilmer:
as if you could somehow have uniqueness without distinctivness

cy:
I have no idea, I think we left it out originally and now there's a ticket about it. Anyway I'll ask bessie.

cy:
she says go with unique. updated

PR #2005, opened 2017-08-10
🐛 [PRTL-1519] fix dropdown not disappearing after dragging

Jephuff:


alex-wilmer:
is this bug specific to a browser? I can't reproduce the bug in the ticket in chrome on nightly / prod

Jephuff:
It happens for me in chrome. (It only happens after dragging and then clicking off the dropdown, clicking on the dropdown will fix it)
I don't think bessie was able to reproduce with her touchpad on the new macbook, so maybe something about the new hardware? Happens for me with my mouse and trackpad.

PR #2007, opened 2017-08-10
♿️ [PRTL-1573, PRTL-1574] modal 508

Jephuff:


Jephuff:
Found it searching `aria-hidden` in their source code :)

PR #2008, opened 2017-08-10
:bug: PRTL-1567 RepoCases, sort by annotation

cy:


alex-wilmer:
interesting. this default sorts cases by # of annotations right? but we didn't have a default sort before. maybe we should get this reviewed by a BA before merging?

cy:
bessie thinks it's good

PR #2010, opened 2017-08-10
next -> master

alex-wilmer:


shanewilson:
these are for the next release (1.9.0) not the current one (1.8.0) right?

alex-wilmer:
sets definitely, probably same with facet refactor. not sure about the other

PR #2015, opened 2017-08-11
♿️ [PRTL-1576] suggestion facet 508

Jephuff:


src/packages/@ncigdc/components/Aggregations/SuggestionFacet.js
```diff
@@ -199,10 +199,14 @@ const SuggestionFacet = compose(
                       onKeyDown={selectableList.handleKeyEvent}
                       placeholder={placeholder}
                       value={inputValue}
-                      aria-activedescendant={_.get(
-                        selectableList,
-                        'focusedItem.id',
-                      )}
+                      aria-activedescendant={
+                        active
```
alex-wilmer: does `undefined` make a difference here vs false / null ?

Jephuff: `null` works 👍 updated

alex-wilmer: Oh i meant why not `active && ...`

Jephuff: Unfortunately false gets turned into a string so it won't work. Has to be either null or undefined.

alex-wilmer: gotcha. can you leave a comment there about that?

alex-wilmer: someone like me might come and erase that later not knowing

Jephuff: definitely! :) 
done.

PR #2019, opened 2017-08-11
PRTL-1579 dont use relay to fetch case file ids for add to cart btn

cy:


src/packages/@ncigdc/components/AddCaseFilesToCartButton.js
```diff
@@ -23,44 +26,79 @@ const styles = {
...
-          isFilteredFileDataRequired: true,
-          filesFilters: filters,
-        },
-        readyState =>
-          _.some([readyState.ready, readyState.aborted, readyState.error]) &&
-          setIsLoading(false),
-      );
+      setFetchedFiles(await fetchFiles(caseId, fileCount));
+      if (filters) {
+        setFetchedFilteredFiles(await fetchFiles(caseId, fileCount, filters));
```
alex-wilmer: that's kind of interesting. is this the same as

```
let something = await promise()
setSomething(something)
```

I guess so. cool 👍

cy: wondered if it would work and it did

PR #2020, opened 2017-08-11
♻️ PRTL-1572 modern FilesTable / remove classic cart route

alex-wilmer:
This PR turns the FilesTable and cart summary sections into modern components. This fixes a bug in the smart search page where one could not sort  anymore, but just for the files table.


alex-wilmer:
breaking up https://github.com/NCI-GDC/portal-ui/pull/2056

PR #2023, opened 2017-08-14
✨ PRTL-1369 handle overlapping lolliplot domains

alex-wilmer:


src/packages/@ncigdc/modern_components/Lolliplot/separateOverlapping.js
```diff
@@ -0,0 +1,18 @@
+export const overlaps = (a, b) => {
+  const first = a.start < b.start ? a : b;
+  const last = first.start === a.start ? b : a;
+  return first.end > last.start;
+};
+
+const separateOverlapping = data => {
+  return data.slice(1).reduce((acc, d) => {
```
Jephuff: Would probably be good to remove the slice and set the initial value to [], I don't think it changes the output at all?

src/packages/@ncigdc/modern_components/Lolliplot/separateOverlapping.js
```diff
@@ -0,0 +1,18 @@
...
+    const index = acc.findIndex(level => level.every(l => !overlaps(l, d)));
+
+    if (index > -1) acc[index].push(d);
+    else acc[acc.length] = [d];
```
Jephuff: Could use push here as well, for consistency.

alex-wilmer:
@Jephuff updated, toggle added

src/packages/@ncigdc/modern_components/Lolliplot/separateOverlapping.js
```diff
@@ -0,0 +1,21 @@
...
+  return data.reduce(
+    (acc, d) => {
+      const index = acc.findIndex(level => level.every(l => !overlaps(l, d)));
+
+      if (index > -1) acc[index].push(d);
+      else acc.push([d]);
+
+      return acc;
+    },
+    [[]],
```
Jephuff: You shouldn't need the empty array in the array :D with `[]` index will equal `-1` so it will push a new array. 
not sure if it matters, but this way if `data === []`  it will return `[[]]` instead of `[]`. `[]` seems more correct? 
(sorry for the nitpicking :P )

alex-wilmer: woops just a mistake

PR #2024, opened 2017-08-15
code split

cy:
errors with ```Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object. Check the render method of `LoadableComponent`.``` without `import-inspector`


cy:
bundle.js went from 2.0mb to 1.3mb in dev mode

cy:
@Jephuff oops, deleted

src/packages/@ncigdc/components/LoadableWithLoading/Loading.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import Overlay from '@ncigdc/uikit/Overlay';
+import Spinner from '@ncigdc/uikit/Loaders/Material';
+
+const Loading = ({ isLoading, timedOut, pastDelay, error }) => {
```
jbarno: timedOut will never be true, you need to pass a value for the timeout to the Loadable object in index.js below.
[npmdocs]( https://www.npmjs.com/package/react-loadable#timing-out-when-the-loader-is-taking-too-long)  [githubdocs4.0.2version](https://github.com/thejameskyle/react-loadable/tree/v4.0.2#optstimeout)

src/packages/@ncigdc/components/LoadableWithLoading/Loading.js
```diff
@@ -0,0 +1,31 @@
...
+      return null;
+    }
+  } else if (error) {
+    // If we aren't loading, maybe
+    return <div>Error! Component failed to load</div>;
+  } else {
+    // This case shouldn't happen... but we'll return null anyways.
+    return null;
+  }
+};
```
jbarno: It would be nice to have unit tests for this function, seems really important as its wrapping all the routes.

src/packages/@ncigdc/components/LoadableWithLoading/Loading.js
cy: didn't set a timeout because I think it's better for a page to eventually load then timeout when it would have eventually resolved...at least for this project, we end up with tickets about timeouts. :woman_shrugging: But we can definitely set one if necessary.

src/packages/@ncigdc/components/LoadableWithLoading/Loading.js
cy: Since this PR is already merged, could you add a ticket for the unit test?

jbarno: https://jira.opensciencedatacloud.org/browse/PRTL-1797

PR #2027, opened 2017-08-15
🐛 PRTL-1489 project, not case, filter for case entity ssm table

alex-wilmer:


alex-wilmer:
This is kind of hackish, I'm passing in a custom total value to the ssms table only in the case entity page. Very open to suggestions for a different strategy.

Jephuff:
image

Jephuff:
I'm a little confused. Isn't the problem that that table should be filtered by case_id and it's currently filtered by project? would passing in defaultFault filters as case_id and then adding a contextFilters for getting the filteredCases work?

alex-wilmer:
>< right I needed to do the inverse of what I did kind of

alex-wilmer:
oh I think it might need more change than that since `score` is being used for that column. maybe instead of passing additional filters, pass in a context component

alex-wilmer:
or maybe even export a slightly modified version of the table model

alex-wilmer:
@Jephuff okay I think I implemented what you suggested and I think this is good now.

phuongmy:
Tested the PR just now and it looks good, great job, thanks Alex!

src/packages/@ncigdc/routes/CaseRoute.js
```diff
@@ -18,16 +18,23 @@ import {
...
-  />,
+const SsmsTable = createCaseSummary(
+  ({
+    viewer,
+    node = viewer.repository.cases.hits.edges[0].node,
+    projectId = node.project.project_id,
+    caseId = node.case_id,
+  }) =>
+    <ST
+      defaultFilters={makeFilter([{ field: 'cases.case_id', value: caseId }])}
```
Jephuff: I think this filter is passed in already on line 110.

alex-wilmer: ah right since I switched the filters around. i'll remove it

PR #2029, opened 2017-08-16
✨ [PRTL-1479] use JSON.stringify instead of JSURL.stringify

Jephuff:


alex-wilmer:
doesn't this change the filters to JSON encoded instead of jsurl encoded though? We were thinking of creating a button that does a copy to curl kind of thing or maybe provided a few options https://oicr-gdc.slack.com/archives/C0H0726P7/p1502727378335314

Jephuff:
Ahh I missed that. Are we attached enough to JSURL to worry about it? I vaguely remember @shanewilson talking about removing it anyways, but I might be misremembering that.

shanewilson:
I just added it to test it out - I don't have a strong opinion on it. If it causes us trouble probably best to remove it entirely.

shanewilson:
But I don't like the idea of some json and some jurl - if we remove it it should be totally gone - no jurl anywhere in the code.

Jephuff:
JSURL is left as a backup for parsing so existing links don't break, though it might be confusing because I named the file jsurl that does the parsing just so I could change less code. I can update that though.

shanewilson:
Deploy preview processing.

Built with commit 1741f455deb5ce4452be74d0f3185889c34357b2

https://app.netlify.com/sites/ncigdc/deploys/599ae9e17960b113123811cb

PR #2038, opened 2017-08-22
✨ PRTL-1601 wrap project vis in HasCases

alex-wilmer:


src/packages/@ncigdc/modern_components/HasCases/index.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import withData from './HasCases.relay';
+export default withData(
+  p =>
+    p.viewer.explore.cases.hits.total ? React.Children.only(p.children) : null,
```
Jephuff: Is there an advantage to forcing the only child here, could just wrap children in a div here rather then where it's used.

alex-wilmer: not sure what you mean. until react 16, all components are always a single child

Jephuff: I mean replace `React.Children.only(p.children)` with `<div>{p.children}</div>` just so who ever it is using can wrap anything with it.

alex-wilmer: when react 16 comes out, you can just return `this.props.children` immediately

alex-wilmer:
@Jephuff updated

PR #2039, opened 2017-08-22
✨ [PRTL-1598] add analysis menu

Jephuff:
Start of the analysis menu. Not sure what the requirements are for each analysis so this will need to be update with those requirements.


alex-wilmer:
I think we should merge this. since a few of us will be working off of it, we can add changes without needing to port them back here. this seems good the way it is for me

PR #2041, opened 2017-08-23
🐛 [PRTL-1604] wait for accept to disable first time popup

Jephuff:


alex-wilmer:
This will keep the popup coming back until they click accept right? The ticket says the modal should only be closed by clicking accept (disable outside click). I think your code makes sense, but I foresee the ticket reappearing. What do you think?

Jephuff:
Yep, I didn't read that ticket carefully enough. Will update!

PR #2042, opened 2017-08-23
:sparkles: PRTL-1597 add Manage Sets page

cy:
screen shot 2017-08-23 at 12 44 08 pm
~~doesn't use the streaming `DownloadButton` component for the dl tsv/json yet trying to figure out why errors here https://github.com/NCI-GDC/gdcapi/blob/0b29587d19c836b0276d98c29b7d2c1eded31ef4/gdcapi/resources/index/v0.py#L33~~ figured it out, needed to update elasticsearchpy locally

### DIFF
--- a/src/packages/@ncigdc/modern_components/Counts/Count.js
+++ b/src/packages/@ncigdc/modern_components/Counts/Count.js
@@ -1,12 +1,17 @@
 // @flow
 import React from 'react';
 import { get } from 'lodash';
-import { compose, withProps } from 'recompose';
+import { compose, withProps, withPropsOnChange } from 'recompose';
 
 export default compose(
   withProps(({ viewer, path }) => {
     return {
-      count: get(viewer, path, 0).toLocaleString(),
+      count: get(viewer, path, 0),
     };
   }),
-)(({ count }) => <span>{count}</span>);
+  withPropsOnChange(['viewer'], ({ count, handleCountChange, loading }) => {
+    if (!loading && handleCountChange) {
+      handleCountChange(count);
+    }
+  }),
+)(({ count }) => <span>{count.toLocaleString()}</span>);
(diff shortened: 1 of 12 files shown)


src/packages/@ncigdc/uikit/EditableLabel.js
```diff
@@ -0,0 +1,72 @@
...
+              if (e.key === 'Enter') {
+                toggleEditing();
+              }
+            }}
+            onBlur={toggleEditing}
+            type="text"
+            autoFocus
+          />
+        : <Row
+            onClick={toggleEditing}
```
Jephuff: Should add either `cursor: pointer` or `cursor: text` to this.

src/packages/@ncigdc/modern_components/Counts/Count.js
```diff
@@ -1,12 +1,17 @@
...
 
 export default compose(
   withProps(({ viewer, path }) => {
     return {
-      count: get(viewer, path, 0).toLocaleString(),
+      count: get(viewer, path, 0),
     };
   }),
-)(({ count }) => <span>{count}</span>);
+  withPropsOnChange(['count'], ({ count, handleCountChange }) => {
```
Jephuff: Could you switch this to use utils/withPropsOnChange, it's the same but won't have an error if setting state during construction.

cy:
tsv downloads for sets of size 10k were timing out, switched to using streaming for those files as well. Depends on https://github.com/NCI-GDC/gdcapi/pull/836

Jephuff:
the remove empty or deprecated button and the warning that a set is empty or deprecated have a little flash on load. Might be a bit of a timing issue?

cy:
@Jephuff thanks fixed it

PR #2047, opened 2017-08-24
✨ memoizee caching for modern env

alex-wilmer:
This isn't default behaviour in modern, just simple hash based caching like varnish for local session.

### DIFF
--- a/src/packages/@ncigdc/modern_components/environment.js
+++ b/src/packages/@ncigdc/modern_components/environment.js
@@ -7,6 +7,7 @@ import { setLoader, removeLoader } from '@ncigdc/dux/loaders';
 
 const source = new RecordSource();
 const store = new Store(source);
+const simpleCache = {};
 const handlerProvider = null;
 
 function fetchQuery(operation, variables, cacheConfig, uploadables) {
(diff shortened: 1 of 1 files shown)


src/packages/@ncigdc/modern_components/environment.js
```diff
@@ -20,9 +21,12 @@ function fetchQuery(operation, variables, cacheConfig, uploadables) {
...
-
   const [componentName] = operation.name.split('_relayQuery');
 
+  if (simpleCache[hash]) {
+    return Promise.resolve(simpleCache[hash]);
+  }
+
   setTimeout(() => reduxStore.dispatch(setLoader(componentName)));
 
   return fetch(
```
Jephuff: Could we use memoizee on this fetch call instead?

alex-wilmer: I'll try it

alex-wilmer: well actually I think it might not be better to do that (unless I'm missing something). the args for fetch might change, but the if the hash is the same then the result should be the same

Jephuff: hmm, the only think that could be different would be `componentName` right? Because relay forces us to have that in the query anyways.

alex-wilmer:
https://github.com/NCI-GDC/portal-ui/pull/2048

PR #2052, opened 2017-08-24
1458 user profile page

cy:
screencapture-localhost-3000-user-profile-1503608081412

### DIFF
--- a/src/packages/@ncigdc/components/Modals/UserProfileModal.js
+++ b/src/packages/@ncigdc/components/Modals/UserProfileModal.js
@@ -1,0 +1,80 @@
+// @flow
+import React from 'react';
+import BaseModal from '@ncigdc/components/Modals/BaseModal';
+import { compose } from 'recompose';
+import { connect } from 'react-redux';
+
+import Hidden from '@ncigdc/components/Hidden';
+import { withTheme } from '@ncigdc/theme';
+import Check from '@ncigdc/theme/icons/Check';
+import { Row } from '@ncigdc/uikit/Flex';
+import EntityPageHorizontalTable from '@ncigdc/components/EntityPageHorizontalTable';
+import withRouter from '@ncigdc/utils/withRouter';
+
+const enhance = compose(
+  connect(({ auth: { user } }) => ({ user })),
+  withTheme,
+  withRouter,
+);
+
+const UserProfileModal = ({
+  user: { username, projects },
+  theme,
+  push,
+  query,
+}) => {
+  const flattenedProjects = Object.keys(projects).reduce(
+    (acc, k) => ({ ...acc, ...projects[k] }),
+    {},
+  );
+  const allValues = [
+    ...new Set(
+      Object.keys(flattenedProjects).reduce(
+        (acc, projectId) => [...acc, ...flattenedProjects[projectId]],
+        [],
+      ),
+    ),
+  ];
+
+  const data = Object.keys(flattenedProjects).map((projectId, i) => ({
+    projectId,
+    ...allValues.reduce(
+      (acc, v) => ({
+        ...acc,
+        [v]: flattenedProjects[projectId].includes(v)
+          ? <span><Check /><Hidden>True</Hidden></span>
+          : <Hidden>False</Hidden>,
+      }),
+      {},
+    ),
+  }));
+  return (
+    <BaseModal
+      title={`Username: ${username}`}
+      closeText="Done"
+      contentStyle={{ padding: 0 }}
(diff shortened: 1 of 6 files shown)


alex-wilmer:
this breaks if you go to that page not logged in. maybe add a branch + redirect?

```
const enhance = compose(
  connect(({ auth: { user } }) => ({ user })),
  branch(({ user }) => !user, renderComponent(() => <Redirect to="/" />)),
);
```

Also could you please create a folder in `routes` with your Route + Page files? I'm slowly clearing out the `containers` folder which was meant for classic relay containers. thanks!

src/packages/@ncigdc/routes/UserProfileRoute/UserProfileRoute.js
```diff
@@ -0,0 +1,66 @@
...
+import React from 'react';
+import { compose, branch, renderComponent } from 'recompose';
+import { connect } from 'react-redux';
+import { Redirect } from 'react-router';
+
+import { Row } from '@ncigdc/uikit/Flex';
+import Hidden from '@ncigdc/components/Hidden';
+import Heading from '@ncigdc/uikit/Heading';
+import Table, { Tr, Td } from '@ncigdc/uikit/Table';
+import CheckedSquareO from '@ncigdc/theme/icons/CheckedSquareO';
```
Jephuff: How would you feel about just a checkmark without the square, and nothing if they don't have it? These make me feel like I can change my permissions haha

cy: haha true

cy:
screencapture-localhost-3000-user-profile-1503688743861

alex-wilmer:
what do you think of using fullwidth layout title for username

image

```
    <FullWidthLayout
      title={'ALEX WILMER'}
      entityType={<i className="fa fa-user" />}
    >
```

cy:
Sure - kinda looks like the case entity page though, guess the icon is 'CA' there. Currently fighting with travis

cy:
screen shot 2017-08-28 at 2 06 18 pm
as a modal. ready for re-review

src/packages/@ncigdc/components/UserDropdown.js
```diff
@@ -50,6 +59,12 @@ const UserDropdown = connect(state => ({
         </NavLink>
       }
     >
+      <DropdownItem onClick={() => dispatch(setModal(<UserProfileModal />))}>
```
Jephuff: Could you add `cursor: pointer` to these?

src/packages/@ncigdc/components/Modals/UserProfileModal.js
```diff
@@ -0,0 +1,135 @@
...
+                display: 'flex',
+                justifyContent: 'center',
+                alignItems: 'center',
+              }}
+            >
+              <SearchIcon size={14} />
+            </div>
+            <Hidden>filter cases</Hidden>
+          </label>
+          <input
```
Jephuff: can you either pass in `searchTerm` as the initial value or reset the query param on open? 
right now if you search for something and reopen the modal the results are filters but the input is empty.

src/packages/@ncigdc/components/Modals/UserProfileModal.js
```diff
@@ -0,0 +1,135 @@
...
+          />
+        </Row>
+        <Row style={{ justifyContent: 'center' }}>
+          <LocalPaginationTable
+            className="test-consequences-table"
+            data={data}
+            prefix={paginationPrefix}
+            entityName="projects"
+            style={{ width: '70%' }}
+            enablePickSize={false}
```
Jephuff: Do we need to disable picking size? 10 is a pretty small number, might be better to just have overflow?

cy: both scrolling and pagination in a modal seems too much and also didn't want to deal with small issues like the picker being under the modal
screen shot 2017-08-30 at 3 17 43 pm
maybe I'll do away with the front end pagination and see what phuong-my thinks of just scroll

shanewilson: the full page you had before seems better

cy: reason for changing to modal was they didn't want it to be a full data page, and they wanted to "have a screen where users can see it without navigating away from current screen"

cy:
@alex-wilmer @Jephuff ready for re-veiw

PR #2055, opened 2017-08-25
♻️ modern ProjectsCharts

alex-wilmer:
requires https://github.com/NCI-GDC/gdcapi/pull/821


alex-wilmer:
1800+ lines are just generated stuff

Jephuff:
bottom spacing is missing in loading state.
image

It also seems weird when changing filters. It updates both graphs but then continues to have the loader?

PR #2060, opened 2017-08-28
✨ [PRTL-1606] add count in dropdown

Jephuff:


alex-wilmer:
image
(disregard my red line)

this is looking pretty crowded. (checkbox overlapped). Maybe we can use a single truncated line for the set name amd have `box | name | # items`  per line. what do you think?

Jephuff:
@alex-wilmer truncated 👍

Jephuff:
Good call, fixed :D

PR #2064, opened 2017-08-29
check localstorage for react envs first

alex-wilmer:
should probably make this work without needing to refresh the page, but just testing it out


alex-wilmer:
> How would you feel about adding a notification/banner that the api has an override. I could see some confusing down the road :slightly_smiling_face:

PR #2071, opened 2017-08-30
♻️ refactor analysisRoute to fix route order bug

Jephuff:


cy:
Something's still wrong 😞, you don't get:
```
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object. Check your code at PortalContainer.js:128.
    in lifecycle(PortalContainer) (created by Connect(lifecycle(PortalContainer))) [...]```

PR #2074, opened 2017-08-31
✨ [PRTL-1600] set operations

alex-wilmer:


alex-wilmer:
https://github.com/NCI-GDC/portal-ui/pull/2089

PR #2080, opened 2017-09-01
✨ [PRTL-1628] limit size on create and append

Jephuff:
requires https://github.com/NCI-GDC/gdcapi/pull/837


cy:
when try append set by checking cases from the table, I get 
screen shot 2017-09-06 at 2 06 38 pm

```
{
  "data": {
    "sets": {
      "append": {
        "explore": {
          "case": null
        }
      }
    }
  },
  "errors": [
    {
      "locations": [
        {
          "column": 9,
          "line": 8
        }
      ],
      "message": "TransportError(400, u'parsing_exception', u'[bool] malformed query, expected [END_OBJECT] but found [FIELD_NAME]')"
    }
  ]
}
```

Jephuff:
@cy fixed it, doesn't like filters not wrapped in `and` apparently.

cy:
yeah that fixed it...seems to work in graphiql without which is super weird  ¯\_(ツ)_/¯

PR #2083, opened 2017-09-05
💄 [PRTL-1625] analysis menu before repo

alex-wilmer:


src/packages/@ncigdc/components/Header.js
```diff
@@ -148,6 +142,12 @@ const Header = compose(
               </ExternalLink>
 
             </li>}
+          <li>
```
Jephuff: You need to jump one more link, thats the old analysis link. Maybe we should delete it?

alex-wilmer:
oops good catch. yeah we should remove it now I think

PR #2084, opened 2017-09-05
dont append /api to AUTH_API from localstorage

cy:


Jephuff:
How would you feel about adding a `REACT_APP_GDC_AUTH_API` variable to localstorage instead of this logic?

cy:
like this?
```Javascript
export const AUTH_API = localStorage. REACT_APP_GDC_AUTH_API ? localStorage. REACT_APP_GDC_AUTH_API : `${AUTH}/api`;
```
because don't want to have to add it https://github.com/NCI-GDC/tungsten/blob/develop/salt/srv/services/portal/bin/build#L20

alex-wilmer:
I think this is fine.. it's just for convenience. there's a probably a better way to handle more cases once we add them

Jephuff:
It would only be for overriding, so you wouldn't need to add it to tungsten. 
```
export const AUTH = localStorage.REACT_APP_GDC_AUTH || process.env.REACT_APP_GDC_AUTH || '';
export const AUTH_API = localStorage.REACT_APP_GDC_AUTH_API || `${AUTH}/api`;
export const API = localStorage.REACT_APP_API || process.env.REACT_APP_API;
```
You would only need to set `REACT_APP_GDC_AUTH_API` in localStorage, because thats what we actually want to override.

cy:
ok updated

PR #2090, opened 2017-09-06
✨ max set name length

Jephuff:


src/packages/@ncigdc/utils/filtersToName.js
```diff
@@ -19,14 +19,14 @@ function getValues(filters, sets) {
 }
 
 const MAX_VALUES = 6;
-export default function({ filters, max = MAX_VALUES, sets }) {
+export default function({ filters, max = MAX_VALUES, sets, length = 100 }) {
```
alex-wilmer: should this `length = 100` be `length =  MAX_SET_NAME_LENGTH` or is that a different 100

Jephuff: 100 here was just a default, it can be removed and I can make it unlimited if it's not passed in.

src/packages/@ncigdc/utils/filtersToName.js
```diff
@@ -19,14 +19,19 @@ function getValues(filters, sets) {
 }
 
 const MAX_VALUES = 6;
-export default function({ filters, max = MAX_VALUES, sets }) {
+export default function({
+  filters,
+  max = MAX_VALUES,
+  sets,
+  length = Infinity,
```
alex-wilmer: lol this is more fun ∞

PR #2085, opened 2017-09-05
🐛 [PRTL-1602] newrelic bugs

Jephuff:

PR #2091, opened 2017-09-06
1632 handle no mutation data

cy:
(clark release) this PR cherry-picks 1601 use HasCases in project entity to not show viz components when no mutation data, and adds a similar HasSsms to use in the Case entity page. ~~The other issue in 1632 ticket about the project entity page not loading when no cases is a separate issue on `next`, 1601 doesn't cause it; will fix separately on next.~~


cy:
so code splitting was breaking HasCase made the whole project page throw `Uncaught (in promise) Error: Attempted to update component `ReactCompositeComponent` that has already been unmounted (or failed to mount).`, fixed by @Jephuff https://github.com/NCI-GDC/portal-ui/pull/2085 on `next`. There's no code splitting on master yet so this branch should be good to go for release.

PR #2098, opened 2017-09-07
Share/analysis

alex-wilmer:


src/packages/@ncigdc/components/analysis/AnalysisResult.js
```diff
@@ -18,6 +18,27 @@ const enhance = compose(
   withRouter,
 );
 
+const CopyButton = withState('copied', 'setCopied', false)(p =>
+  <Clipboard
```
Jephuff: Might be a good idea to provide the text in a way that the user can select/and copy and paste the old fashion way as well. That seems typical and not all browsers handle this as nicely as chrome :D

PR #2104, opened 2017-09-11
🐛 [PRTL-1646, PRTL-1645] ssm -> mutation and add missing ssm append button

Jephuff:


alex-wilmer:
I think I need this fix before https://github.com/NCI-GDC/portal-ui/pull/2109 which is a very small change. @Jephuff  maybe add it here?

PR #2109, opened 2017-09-11
✏️ [PRTL-1647, PRTL-1716, PRTL-1715] no limit options in append to set

alex-wilmer:


Jephuff:
We should probably keep the warning just so the user is aware of the truncating. Checked with @phuongmy and she is ok with it.

phuongmy:
Agree. But only if the user tries to add more than 50K item to the set!

alex-wilmer:
image

shanewilson:
how does that set have 50,002 already?
shouldn't the message says that 68,768 won't be saved?

Jephuff:
You can save in 50,000 chunks (like icgc)

src/packages/@ncigdc/modern_components/CohortComparison/__generated__/CohortComparison_relayQuery.graphql.js
```diff
@@ -0,0 +1,389 @@
+/**
```
Jephuff: Could you delete this file?

src/packages/@ncigdc/components/Modals/AppendSetModal.js
```diff
@@ -41,61 +40,74 @@ const AppendSetModal = ({
...
-
+      {selected &&
+        <Count>
+          {count =>
+            <span>
+              {total + count >= max &&
+                <WarningBox>
+                  Above maximum of {pluralize(displayType, max, true)}.
+                  {count < max &&
+                    <span>&nbsp;Only {max - count} will be saved.</span>}
```
Jephuff: Didn't we talk about checking for duplicate ids? or did that not work out?

alex-wilmer: sorry I'm a little fuzzy on the details of that conversation. You mean to make sure that only items which are not already in the set are added, right?

Jephuff: more making it so it doesn't count duplicate ids against the total. 
so I believe right now if you were to save 30,000 and then append the same 30,000 it would tell you it's being truncated when really there are only 30,000 unique ids.
only really an issue if the append also contains 1 new id but then is not added because of the truncating.

PR #2118, opened 2017-09-12
1599 cohort pvalue

shanewilson:


src/packages/@ncigdc/modern_components/Pvalue/Survival.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
```
Jephuff: This file isn't used anywhere?

shanewilson: oh yea copy/pasted a directory to get started

PR #2119, opened 2017-09-13
:sparkles: PRTL-1657 allow upload set in manage sets

cy:


shanewilson:
👍 works! maybe the `Upload Set` button could have a little down arrow to show it is a drop down.

cy:
updated, also was copying the down arrow from the cart and noticed that the table header was blocking the buttons 😵, pushed the table down.

PR #2120, opened 2017-09-13
✨ [PRTL-1656] add venn to cohort comparison

Jephuff:


shanewilson:
looks good. minor note - hovering over the venn diagram still looks like you can interact with it.

Jephuff:
I'm adding a click handler to the chart right now to create the set operation analysis. Might still be confusing that you can click on the venn here but it does something different? Do you think the link should only be on text under the Venn?

PR #2124, opened 2017-09-15
⬆️ npm install with npm 5.4

alex-wilmer:


Jephuff:
Shouldn't this be into next?

alex-wilmer:
Yes & I need to downgrade my atom packages as to not have prettier mess up this code or:

No & do it here and merge this into next since we'll need it there anyways

--- 

can be done in either order

Jephuff:
hmm, concerned with running lint on master, could cause some real issues when rebasing next. next also seems to not have any changes when running `npm i`? could we hold off until next is merged into master?

alex-wilmer:
sure

Jephuff:
Or just drop the `eslint --fix` commit, package-lock.json conflicts are easy enough to resolve :)

alex-wilmer:
but then travis fails on a zillion eslint errors. I'll just rollback my editor

alex-wilmer:
https://github.com/NCI-GDC/portal-ui/pull/2127

PR #2125, opened 2017-09-15
✨ [PRTL-1659] save set name in analysis

Jephuff:


src/packages/@ncigdc/components/analysis/SelectSet.js
```diff
@@ -84,14 +86,20 @@ const SetTable = ({
...
-                  const sets = ids.length
-                    ? {
-                        ...selectedSets,
-                        [type]: ids,
-                      }
-                    : omit(selectedSets, type);
-                  setSelectedSets(sets);
+                  const typeSets = selectedSets[type] || {};
+                  const setId = e.target.value;
+                  const newTypeSet = typeSets[setId]
```
alex-wilmer: took me a minute to understand what the change was here, but I see it now. Do you think you could add a comment explaining what 'typeSet' is in this context?

Jephuff: Took me a second to understand too, I'll take a stab at making it clearer 👍

Jephuff: Refactored so the on change looks like 
```
onChange={e => {
  const setId = e.target.value;
  const setIdPath = [type, setId];
  setSelectedSets(
    _.get(selectedSets, setIdPath)
      ? removeEmptyKeys(_.omit(selectedSets, setIdPath))
      : _.set({ ...selectedSets }, setIdPath, sets[setId]),
  );
}}
```
A bit clearer, what do you think?

alex-wilmer: definitely clearer 👍

PR #2126, opened 2017-09-15
:sparkles: PRTL-1655 download all sets in a tar

cy:
requires https://github.com/NCI-GDC/gdcapi/pull/845. 

A tar with all the cases, all the genes and 50K mutations takes 58s to start streaming. Can't start streaming right away because the TarStream function needs to know the full contents size before it starts, otherwise tar will be incomplete/garbage.

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -592,42 +592,61 @@ type CaseAggregations {
   case_id: Aggregations
   samples__sample_type_id: Aggregations
   samples__updated_datetime: Aggregations
-  samples__biospecimen_anatomic_site: Aggregations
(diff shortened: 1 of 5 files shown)


Jephuff:
Could we also make it export selected instead of all to match behaviour of the delete button?

cy:
yeah, the notification's not going away because the domain name's different localhost vs. collab api; it'll go away when live. The api removes a cookie when it's done that the checker looks at.

cy:
Updated, and I actually forgot to remove the token from the cookie on the api so fixed it up there as well

PR #2129, opened 2017-09-18
Fix/pvalue nullcrash

alex-wilmer:
- fix pvalue and survival null crashes when buckets return null.
- pass initial value to InputWithWarning

### DIFF
--- a/src/packages/@ncigdc/modern_components/CohortComparison/CohortComparison.js
+++ b/src/packages/@ncigdc/modern_components/CohortComparison/CohortComparison.js
@@ -170,16 +170,29 @@ export default compose(
           <Survival
             survivalData={survivalData}
             result1={result1}
+            Alias={Alias}
             result2={result2}
             set1id={set1}
             set2id={set2}
             palette={[set1_colour, set2_colour]}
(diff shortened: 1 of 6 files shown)


src/packages/@ncigdc/modern_components/CohortComparison/CohortComparison.js
```diff
@@ -176,9 +176,7 @@ export default compose(
             palette={[set1_colour, set2_colour]}
             style={{ flex: 1, width: 100 }}
           />
-          <div
-            style={{ marginLeft: 50, flex: 1, width: 100, textAlign: 'center' }}
-          >
+          <div style={{ marginLeft: 50, flex: 1, width: 100 }}>
```
Jephuff: the link should probably still be centered
image

alex-wilmer: how about this instead:

image

Jephuff: works for me 👍

Jephuff:
Could you make the link inline-block? Wraps weird right now 
image

PR #2132, opened 2017-09-19
✨ [PRTL-1669] demo sets

Jephuff:
requires https://github.com/NCI-GDC/gdcapi/pull/847


cy:
Not sure if it's intentional that the demo sets aren't saved in redux store, but that they're not stored causes some strange behaviour like pressing this 'open in a new tab' 
screen shot 2017-09-25 at 11 22 25 am
the current tab disappears, but this new analysis is saved. And when there is a list of saved analyses, and you are on a demo analysis, clicking on one of the saved ones causes the demo tab to disappear. Maybe don't add the demo analyses to the tabs?

Jephuff:
Hmm, my intended behaviour was it wouldn't disappear until you left the analysis results tab or clicked the  "x" beside it, something must have changed. Do you think thats a better behaviour, or is not showing up in the tabs at all make more sense?

cy:
Yeah, I think disappearing is strange. Either not showing up in the tabs at all, or save them in the store.

Jephuff:
@cy ready again, the demo isn't added to tabs now.

PR #2143, opened 2017-09-28
🐛 [PRTL-1685] fix titles

Jephuff:


src/packages/@ncigdc/components/GitHut/columns.js
```diff
@@ -58,7 +58,7 @@ export default function columns(params) {
     },
     {
       id: 'primary_site',
-      displayName: ['Primary', 'Site'],
+      displayName: ['Major', 'Primary', 'Sites'],
```
cy: haha this is so strange

Jephuff: It sure is haha

PR #2146, opened 2017-10-02
✨ [PRTL-1684] graph sizes

Jephuff:


cy:
make the venn a tad bigger? The intersection part is quite tight
screen shot 2017-10-10 at 11 50 09 am

Jephuff:
@cy a little more room :)
image

PR #2156, opened 2017-10-10
✨ [PRTL-1728] disable sets if above max

Jephuff:


src/packages/@ncigdc/components/SetTable.js
```diff
@@ -33,29 +51,49 @@ const SetTable = ({
   type,
   field,
   style,
+  counts,
+  setCounts,
+  getDisabled = () => null,
```
cy: this function name is a little mysterious, feel like it should return a bool but it returns a string. Maybe rename to `getDisabledMessage`

Jephuff: renamed :)

PR #2162, opened 2017-10-12
⬆️ react@16

alex-wilmer:
- bring sortable into codebase
 - update deps / small changes for 16


alex-wilmer:
some warnings / crashes need to be fixed still

PR #2164, opened 2017-10-13
wip hide portal behind login screen via ENV

alex-wilmer:


shanewilson:
does the AWG portal need to show a different set pages? or was that requirement removed?

PR #2174, opened 2017-10-25
Correct biospec deep link type

cy:
Noticed this bug while working on the slide images, cherry picking the fix over. When deep linking to the biospec tree, the type was always set to 'Sample' so the vertical table headings were wrong.

Also react-highlighter seems to be crapping out, replace with InternalHighlight from quicksearch, might want to move that out and wrap as a component.


cy:
seems like we're merging the image viewer so i'll close this

PR #2176, opened 2017-10-27
CohortComparision tweaks

cy:
- use percentage in y axis
- only use exclusive case sets for survival, requires https://github.com/NCI-GDC/gdcapi/pull/857


alex-wilmer:
This was approved by vincent by the way, so merge away

PR #2183, opened 2017-10-30
1763 demo tab

cy:
~~wip~~ done

### DIFF
--- a/src/packages/@ncigdc/components/analysis/availableAnalysis.js
+++ b/src/packages/@ncigdc/components/analysis/availableAnalysis.js
@@ -142,18 +142,6 @@ const availableAnalysis: Array<TAnalysis> = [
       },
       type: 'set_operations',
     },
-    DemoComponent: props => {
-      const type = ['case', 'gene', 'ssm'].find(t => props.sets[t]);
-      return (
-        <Demo {...props}>
-          <SetOperations
-            type={type}
-            sets={props.sets[type]}
-            message={props.message}
-          />;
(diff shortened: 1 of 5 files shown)


src/packages/@ncigdc/components/analysis/availableAnalysis.js
```diff
@@ -238,9 +235,14 @@ const availableAnalysis: Array<TAnalysis> = [
     setTypes: ['case'],
     validateSets: sets =>
       ['case'].every((t: any) => Object.keys(sets[t] || {}).length === 2),
-    ResultComponent: ({ sets }) => {
-      return <CohortComparison sets={sets} />;
-    },
+    ResultComponent: props =>
```
alex-wilmer: what are your thoughts on using `branch` instead of ternaries?

```
branch(
  props => props.id.includes('demo-'),
  renderComponent(props => (
    <Demo {...props}>
      <CohortComparison sets={props.sets} message={props.message} />
    </Demo>
  )),
)(props => <CohortComparison sets={props.sets} message={props.message} />)
```

cy: what's the benefit... ternaries are language level feature, no need to introduce complexity imo

alex-wilmer: I think the main benefit is separating conditional logic from the main component. but definitely adds noise. 

ok carry on!

PR #2185, opened 2017-10-31
🐛 [PRTL-1848] keep location after uploading set

alex-wilmer:


src/packages/@ncigdc/components/Modals/UploadSet/CreateSetButton.js
```diff
@@ -35,10 +35,8 @@ export default enhance(
         onComplete={setId => {
           onClose();
           push({
-            pathname: '/exploration',
             query: {
               ...query,
-              searchTableTab: `${type}s`,
```
cy: should probably keep this line? It's good feedback if the tab flips to the type of entity uploaded, the user can confirm that the upload succeeded at a glance.

alex-wilmer: This is true. @shanewilson suggested that it should behave like any other facet does, which doesn't have that behaviour (causing the table to flip tabs). I tend to agree, but see your point as well.

cy: hmm fair enough

PR #2197, opened 2017-11-06
:bug: PRTL-1859 download access modal: pass in auth not open files

cy:
fix CartDownloadDropdown.js:L92, should pass authorized files not open files


src/packages/@ncigdc/components/CartDownloadDropdown.js
```diff
@@ -49,16 +49,7 @@ const styles = {
 };
 
 const downloadCart = (user, files, dispatch, setState) => {
```
alex-wilmer: can this be changed to keywords while you're here ?

cy: updated

PR #2201, opened 2017-11-06
✏️ [PRTL-1838] update demo queries

alex-wilmer:
so I'm pretty sure this code fixes the issue, but it looks like the set ids which are used in the demo have been pre made with the wrong filters.

### DIFF
--- a/src/packages/@ncigdc/components/analysis/Demo.js
+++ b/src/packages/@ncigdc/components/analysis/Demo.js
@@ -22,7 +22,11 @@ class Demo extends React.Component {
   }
   componentWillUpdate(nextProps, nextState) {
     nextState.setStates.forEach(({ count, id, type }, i) => {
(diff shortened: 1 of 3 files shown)


cy:
I might be wrong, is https://github.com/NCI-GDC/portal-ui/blob/d585048899510c22e6e89a4fff503215b17a929d/src/packages/%40ncigdc/components/analysis/Demo.js#L26 supposed to create the set with the filters from availableAnalysis?

cy:
https://github.com/NCI-GDC/portal-ui/pull/2201/files#diff-33895a36722d8309c36e182dca686c4bR64 the filter changed, should update the set name to `Bladder, High Impact, Mutect2` to match

alex-wilmer:
updated

alex-wilmer:
hmm I do on netlify but not locally. looking into it

alex-wilmer:
fixed. we weren't passing `filters` along when you click the demo button

PR #2211, opened 2017-11-13
:bug: PRTL-1884 update dispatched close banner action

cy:


src/packages/@ncigdc/dux/notification.js
```diff
@@ -3,10 +3,20 @@ const NOTIFY = 'NOTIFY';
...
+  action: string | null,
+};
+export type TState = TNotification & {
+  closed: boolean,
+};
 
-const notify = payload => ({ type: NOTIFY, payload });
-const closeNotification = payload => ({ type: CLOSE_NOTIFICATION, payload });
+const notify = (payload: TNotification) => ({ type: NOTIFY, payload });
+const closeNotification = (closed: boolean) => ({
```
alex-wilmer: Maybe this doesn't need a payload at all. it looks like every caller is just passing true.

cy: true very true

PR #2214, opened 2017-11-13
🐛 [PRTL-1754] removed orphaned graph componenets if there no unique cases on projects page

aishmitk:


src/packages/@ncigdc/modern_components/ProjectsTable/ProjectsTable.js
```diff
@@ -78,16 +78,15 @@ export default compose(
...
-                        ) : (
-                          <Td key={x.id} />
-                        ),
-                    )}
+                  {tableInfo.filter(x => x.td).map(x => {
+                    if (x.total) {
+                      return <x.total key={x.id} hits={hits} />;
+                    }
+                    if (x.id === 'project_id') {
+                      return <Td key="project_id">Total</Td>;
```
alex-wilmer: is there a ticket to add this? I don't think this will make sense if you start moving the columns around

cy: is it possible to detect the id of the first displayed column and use that in the check/key

alex-wilmer: but what if the first column needs to display an actual total like putting number of cases first. I think we should do something less hacky, not sure what though

PR #2217, opened 2017-11-14
✨[PRTL-1842] More obvious that there is no data for one or both sets in survival plot

aishmitk:


src/packages/@ncigdc/modern_components/CohortComparison/Survival.js
```diff
@@ -124,93 +116,104 @@ export default compose(
...
-                    {survivalData.rawData.results[1] &&
-                      (survivalData.rawData.results[1].donors.length /
-                        result2.hits.total *
-                        100
-                      ).toFixed(0)}%
-                  </Td>
-                </Tr>
-              </tbody>
-            }
+      {get(survivalData, 'rawData.results[0].donors.length', 0) ||
```
cy: add `loading ||` here to fix the flash of 'No survival data' when it's loading

PR #2220, opened 2017-11-17
:bug: PRTL-1888 fix large set id downloads

cy:
for large sets, sometimes the streamed tsv file cut off without all the ids, but the tar_set endpoint which skips the filter and downloads sets directly always worked. Changed the tar_set endpoint to not tar if only one set in payload https://github.com/NCI-GDC/gdcapi/pull/868

done in manage set page, analysis page pending

### DIFF
--- a/src/packages/@ncigdc/components/analysis/SetOperations/OpsTable.js
+++ b/src/packages/@ncigdc/components/analysis/SetOperations/OpsTable.js
@@ -1,18 +1,19 @@
 import React from 'react';
 import { connect } from 'react-redux';
+import { compose } from 'recompose';
+import urlJoin from 'url-join';
+import { AUTH_API } from '@ncigdc/utils/constants';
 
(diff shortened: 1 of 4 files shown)


src/packages/@ncigdc/components/analysis/SetOperations/OpsTable.js
```diff
@@ -32,93 +33,117 @@ const ActionsTd = compose(
...
-            filters={filters}
-            onComplete={setId => {
-              push({
-                pathname: '/repository',
-                query: {
-                  filters: stringifyJSONParam({
-                    op: 'AND',
-                    content: [
+      <Td style={{ textAlign: 'right' }}>
+        {hide ? (
```
alex-wilmer: i'm not normally a fan of multiple returns but I think because it's wrapping such a big chunk of js might be easier to follow if

`if (hide) return <Td ... />`

cy: changing it to `!hide &&`

PR #2223, opened 2017-11-20
✅ PRTL-1797 snapshot testing for Loading

cy:
Well that was fun and easy. Snapshot tests @jbarno


jbarno:
Nice! I think snapshots will be really good for error things, this looks neat.

PR #2227, opened 2017-11-29
✨[PTRL-1376] Added a way to see collisions on lolliplot

aishmitk:


alex-wilmer:
This needs a little more work. going to push this into the next sprint

PR #2229, opened 2017-11-30
1404 sift polyphen update

cy:
~~wip~~ done, requires using `doorman` api


cy:
thanks made those into components, the bubbles alignment, was told to left align them by phuong-my

alex-wilmer:
ok 👍

PR #2230, opened 2017-11-30
Fix/1897 select all checkbox

aishmitk:


src/packages/@ncigdc/tableModels/utils.js
```diff
@@ -109,9 +109,9 @@ export const createSelectColumn: TCreateSelectColumn = ({
     downloadable: false,
     hidden: false,
     th: ({ nodes, selectedIds, setSelectedIds }) => {
-      const ids = nodes.map(node => node[idField]);
+      //TODO: make it easier to understand the structure of nodes obj
```
alex-wilmer: maybe instead:

`// NOTE: "nodes" is really "edges" in the graphql schema`

src/packages/@ncigdc/tableModels/utils.js
```diff
@@ -109,9 +109,10 @@ export const createSelectColumn: TCreateSelectColumn = ({
     downloadable: false,
     hidden: false,
     th: ({ nodes, selectedIds, setSelectedIds }) => {
-      const ids = nodes.map(node => node[idField]);
+      // NOTE: "nodes" is really "edges" in the graphql schema
```
jbarno: Feels like a case to use FlowTypes no? -Not sure if that is possible

src/packages/@ncigdc/tableModels/utils.js
```diff
@@ -109,9 +109,10 @@ export const createSelectColumn: TCreateSelectColumn = ({
...
+      console.log(nodes);
```
jbarno: Debugging log?

src/packages/@ncigdc/tableModels/utils.js
jbarno: Or reaname it edges?

src/packages/@ncigdc/tableModels/utils.js
aishmitk: ah my bad

PR #2231, opened 2017-12-01
✨[PTRL-1909] no saving empty set

aishmitk:


src/packages/@ncigdc/components/Modals/SaveSetModal.js
```diff
@@ -72,9 +72,10 @@ const SaveSetModal = ({
           forceCreate
           forceClick={submitted}
           disabled={
-            !inputName ||
-            inputTotal > max ||
-            inputName.length > MAX_SET_NAME_LENGTH
+            inputTotal.toString() === '' ||
```
cy: casting number to string to check for existence is strange, the problem is the input field returns an empty string when the user deletes everything - so instead, check for that and don't allow the empty strings to be set. And set min to 1 to not allow them to type 0.
ie change https://github.com/NCI-GDC/portal-ui/blob/1f0632ef6c416411d3d72c756a444d58e66d327f/src/packages/%40ncigdc/components/Modals/SaveSetModal.js#L107 to
```Javascript
        <input
          type="number"
          max={max}
          min="1"
          value={inputTotal}
          onChange={e => e.target.value !== '' && setInputTotal(e.target.value)}
        />
```

PR #2235, opened 2017-12-07
1404 sift polyphen update option 2

cy:
screen shot 2017-12-07 at 5 20 28 pm
collapsible impacts, uniqued per impact type


cy:
nor this one

PR #2236, opened 2017-12-08
1404 sift polyphen update with hidden filter

cy:
prob wouldn't use, just to compare


cy:
not using this option

PR #2237, opened 2017-12-08
PRTL-1911 - changes save set modal suggested name to Custom <type> Selection

anncatton:


src/packages/@ncigdc/components/SetActions.js
```diff
@@ -88,6 +88,7 @@ export default enhance(
                     type={type}
                     displayType={displayType}
                     CreateSetButton={CreateSetButton}
+                    selectedIds={selectedIds}
```
cy: looks like `SaveSetModal` takes `setName` as a prop so you could pass `setName={(selectedIds || []).length ? `Custom ${displayType} Set` : '' }` here instead and avoid having to additionally pass selectedIds to `SaveSetModal`? Doesn't super matter but it's less code, and keeps SaveSetModal having to know about selected ids, which are already in the filters.

src/packages/@ncigdc/components/Modals/SaveSetModal.js
```diff
@@ -26,9 +26,13 @@ const enhance = compose(
   withState(
     'inputName',
     'setInputName',
-    ({ filters, displayType, sets, setName }) =>
+    ({ filters, displayType, sets, setName, selectedIds }) =>
```
cy: don't need `selectedIds` anymore

alex-wilmer:
cool this lgtm. @cy what do you think of the types no part of the fn signature?

https://github.com/NCI-GDC/portal-ui/blob/a07f8d02de424f92f4e0633ab4c30ac1ed5df074/src/packages/%40ncigdc/utils/filtersToName.js#L48

cy:
yeah, the file's missing the `/* @flow */` annotation at the top though, was confused why it wasn't causing errors because TFiltersToName should be 
```javascript
type TFiltersToName = ({
    filters: ?Object,
    max?: number,
    sets: Object,
    length?: number,
  }) => string;
```
(it takes 1 destructured object as param not 4 params) and max is assigned to MAX_VALUES, but MAX_VALUES's the thing that's being referenced after and max isn't used

anncatton:
Ah yeah you're right about max.
@alex-wilmer For the joined declaration, is this similar what you talked about with total and MAX_VALUES being inferred?

PR #2241, opened 2017-12-13
:bug:PRTL-1920. resets query so result dropdown disappears on blur.

anncatton:
Wondering why I need to still call setIsInSearchMode(false); it looks like reset() should do this automatically?


cy:
works! Can you change the base branch to `master` actually?

PR #2253, opened 2018-01-03
PRTL-1928. Fix/facets dont reset

anncatton:


alex-wilmer:
@cy CaseFacetSelection does use relay, it's wrapped in a Query. It probably should be generalized but I suggested to do this just for one index type to get started with.

Once we're sure this PR didn't break anything, we should use it as a guide for the other indices.

cy:
OK but actually is it necessary to change custom facets to fix the reset? Looking at prod, the explore case aggregation doesn't have the issue & it's using the old code. If the intention is to move to relay modern for the custom facets then sure, but let's not merge this until the other indices are done and end up in a weird in-between-state on `next`

alex-wilmer:
Sure sounds good

src/packages/@ncigdc/modern_components/FileAggregations/FileAggregations.js
```diff
@@ -0,0 +1,272 @@
...
+        }
+        relay={props.relay}
+        style={{ borderBottom: `1px solid ${props.theme.greyScale5}` }}
+        additionalProps={facet.additionalProps}
+      />
+    ))}
+  </div>
+);
+
+// export const FileAggregationsQuery = {
```
cy: ✂️

PR #2262, opened 2018-01-23
Feat/download clin bio data

anncatton:


anncatton:
I added in new filenames with specific date formats and file extensions where files are downloaded, because we removed that function from the server to prevent double-naming on the new clinical/bio tsv downloads. I thought about making a function for this but I felt it would be more obvious what the name would be if the filename prop was just passed a custom string. Any thoughts?

src/packages/@ncigdc/components/RemoveFromCartSingle.js
```diff
@@ -0,0 +1,33 @@
+/* @flow */
+
+import React from 'react';
+import Button from '@ncigdc/uikit/Button';
+import { Tooltip } from '@ncigdc/uikit/Tooltip';
+import { compose } from 'recompose';
```
alex-wilmer: a little nitpicky, but can you order node_module imports above local ones?

src/packages/@ncigdc/components/RemoveFromCartSingle.js
```diff
@@ -0,0 +1,33 @@
...
+  margin: '0 auto',
+  padding: '0px 5px',
+  ':hover': {
+    background:
+      'linear-gradient(to bottom, #ffffff 50%, #e6e6e6 100%) repeat scroll 0 0 #E6E6E6',
+    borderColor: '#ADADAD',
+  },
+};
+
+export default compose(connect(state => state.cart))(({ dispatch, file }) => (
```
alex-wilmer: it's good practive forward props along, like style etc

```
({ dispatch, file, style = {}, ...props }) => ...
```

so users of this component have some control

src/packages/@ncigdc/modern_components/DownloadBiospecimenDropdown/DownloadBiospecimenDropdown.relay.js
```diff
@@ -0,0 +1,41 @@
...
+  )((props: Object) => {
+    return (
+      <Query
+        parentProps={props}
+        style={{ width: 'auto' }}
+        variables={props.variables}
+        Component={Component}
+        query={graphql`
+          query DownloadBiospecimenDropdown_relayQuery(
+            $filters: FiltersArgument
```
alex-wilmer: This query doesn't really seem to match the name. You're just trying to get the total number of cases right?

src/packages/@ncigdc/modern_components/DownloadClinicalDropdown/DownloadClinicalDropdown.relay.js
```diff
@@ -0,0 +1,41 @@
...
+        variables={props.variables}
+        Component={Component}
+        style={{ width: 'auto' }}
+        query={graphql`
+          query DownloadClinicalDropdown_relayQuery($filters: FiltersArgument) {
+            viewer {
+              repository {
+                cases {
+                  hits(first: 0, filters: $filters) {
+                    total
```
alex-wilmer: same as above. looks like you could reuse the same Query right? check out this file https://github.com/NCI-GDC/portal-ui/blob/next/src/packages/@ncigdc/modern_components/Counts/repositoryCase.relay.js

PR #2272, opened 2018-02-05
🐛 PRTL-1991. Fix clinical entity table width when one tab result.

anncatton:


src/packages/@ncigdc/uikit/SideTabs.js
```diff
@@ -10,11 +10,13 @@ const SideTabs = ({
   activeTab,
   tabs,
   tabContent,
+  hideTabs,
```
cy: instead of passing down hideTabs can you detect tabs.length inside here and not render instead?

anncatton: so, like:
tabs.length && (
    ...render tabs
)
in SideTabs.js

cy: oh how about where you're doing display:none, https://github.com/NCI-GDC/portal-ui/blob/02dc0a8a6eca28046217fd9f796cd81df4ff77dd/src/packages/%40ncigdc/uikit/Tabs.js#L113 instead of passing down `hideTabs` check tabs.length

anncatton: Yeah that's better! But maybe just check
<Column style={!tabs ? { display: 'none' } : {}}>
Just because in this case we're getting rid of tabs if there's only one, but there must be other places where we want to display just one tab?

anncatton: Actually I can just pass in tabs=[] instead of tabs = null from ClinicalCard

PR #2282, opened 2018-03-19
✨ Feat/PRTL-2048 add image viewer file page

anncatton:


src/packages/@ncigdc/modern_components/ImageViewer/ImageViewer.relay.js
```diff
@@ -56,34 +56,32 @@ export default (Component: ReactClass<*>) =>
...
+                    value: ['Slide Image'],
                   },
-                  {
-                    op: 'in',
-                    content: {
-                      field: 'files.data_type',
-                      value: ['Slide Image'],
-                    },
+                },
+                {
```
cy: can you only add the fileId filter if there's a fileId? So that no filter shows all cases with slide image files

src/packages/@ncigdc/components/File.js
```diff
@@ -45,6 +49,67 @@ const styles = {
   },
 };
 
+export const getSlides = caseNode => {
```
cy: rename to `getSlide`, using _.head would only ever have one

PR #2285, opened 2018-03-20
💄Fix/PRTL-2060 image viewer ui feedback

anncatton:


src/packages/@ncigdc/components/ZoomableImage.js
```diff
@@ -40,11 +40,12 @@ const enhance = compose(
               id: 'osd1',
               prefixUrl:
                 'https://cdn.jsdelivr.net/npm/openseadragon@2.3/build/openseadragon/images/',
-              defaultZoomLevel: 1,
+              defaultZoomLevel: 0.9,
```
cy: actually could you put in a comment explaining why 0.9 for future us

PR #2286, opened 2018-03-20
🚸Fix/PRTL-2055 filter cases no slide image

anncatton:


anncatton:
Merged into https://github.com/NCI-GDC/portal-ui/pull/2282

PR #2294, opened 2018-03-23
2078 move view images

cy:


cy:
wait wip

cy:
@anncatton ready for review! Had to switch the filter to looking at summary.experimental_strategy after moving to the 'View Images' button to repo because with the file facet selection panel in the repo page it became really apparent that only files with data_category Biospecimen had data_type slide, although some cases had files of both types; so that the data_category facet looked like it didn't work.

PR #2303, opened 2018-04-04
🚧- No slides in json tooltip

anncatton:
Show slides in excluded tooltip until slide file count can be added to json download

### DIFF
--- a/src/packages/@ncigdc/modern_components/ExploreCasesTable/ExploreCasesTable.js
+++ b/src/packages/@ncigdc/modern_components/ExploreCasesTable/ExploreCasesTable.js
@@ -84,7 +84,7 @@ export default compose(
             arrangeColumnKey="exploreCases"
             total={cases.hits.total}
             endpoint="case_ssms"
-            downloadTooltip="Export All Except #Mutations and #Genes"
+            downloadTooltip="Export All Except #Mutations, #Genes and Slides"
             currentFilters={filters}
             score={score}
             sort={sort}


src/packages/@ncigdc/modern_components/ExploreCasesTable/ExploreCasesTable.js
```diff
@@ -84,7 +84,7 @@ export default compose(
             arrangeColumnKey="exploreCases"
             total={cases.hits.total}
             endpoint="case_ssms"
-            downloadTooltip="Export All Except #Mutations and #Genes"
+            downloadTooltip="Export All Except #Mutations, #Genes and Slides"
```
alex-wilmer: why do Mutations and Genes have a `#`, but Slides does not?

alex-wilmer: What makes you sure it must `literally be "number of"`?

No being sceptical, genuinely curious

anncatton: Hehe it just occurred to me when I saw your comment. But not actually based in logic particularly...tbh I'm not sure why those columns would have it and not the others

cy: this is a hilarious convo :P Reason for the # will have to ask @phuongmy

PR #2304, opened 2018-04-06
✨ PRTL-2081-search-for-bio-type

anncatton:
Also adds styling to cart buttons on case summary and biotree details.

### DIFF
--- a/src/packages/@ncigdc/components/AddToCartButtonAll.js
+++ b/src/packages/@ncigdc/components/AddToCartButtonAll.js
@@ -21,7 +21,6 @@ import Dropdown from '@ncigdc/uikit/Dropdown';
 import DropdownItem from '@ncigdc/uikit/DropdownItem';
 import Button from '@ncigdc/uikit/Button';
 import ShoppingCartIcon from '@ncigdc/theme/icons/ShoppingCart';
-import { iconButton } from '@ncigdc/theme/mixins';
 
 import styled from '@ncigdc/theme/styled';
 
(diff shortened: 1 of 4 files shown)


src/packages/@ncigdc/components/AddToCartButtonAll.js
```diff
@@ -89,14 +89,15 @@ const AddToCartButtonAll = ({
...
-            active={inCart}
             onClick={() =>
               inCart
                 ? dispatch(toggleFilesInCart(edges))
                 : dispatch(addAllFilesInCart(edges))}
             aria-label="Add files to cart"
             role="button"
-            style={iconButton}
+            style={
+              (iconButton, { color: inCart ? '#3c763d' : 'rgb(0, 80, 131)' })
```
cy: just delete iconButton, unless you meant to spread it into style? But seems like you don't need it.
image

src/packages/@ncigdc/modern_components/BiospecimenCard/BioTreeView.js
```diff
@@ -25,8 +31,13 @@ const BioTreeView = ({
   query,
   childrenExpanded,
 }) => {
+  const entityTypes = ['portions', 'analytes', 'aliquots', 'slides'];
   const expanded =
-    ex || (query && entities.hits.edges.some(e => search(query, e).length));
+    ex ||
+    (query &&
+      (entities.hits.edges.some(e => search(query, e).length) ||
+        entityTypes.find(t => t.includes(query)) ||
```
cy: this is cool could you lower case query so it works for `Slides` as well as `slides`

PR #2305, opened 2018-04-09
♿️ - PRTL-2091 - landmark issues

anncatton:


cy:
instead of changing the header/footer elements to plain divs, if you move around the aria roles so they're not nested inside each other axe will be happy https://github.com/NCI-GDC/portal-ui/commit/e2abbd113b32319f4189f9526ed53e44428159d5

PR #2318, opened 2018-06-05
Update Dockerfile for use in Kubernetes

rjackson90:


singergr:
This branch has fallen pretty far out of sync. We can revisit and reopen if needed.

PR #2323, opened 2018-06-19
PRTL-2109 remove file_state references

cy:


DrSpez:
@cy can we merge it?

cy:
@DrSpez merged

PR #2327, opened 2018-06-27
:bug [PRTL-2083] Disable the GoLink Button when the input value is empty.

yuanhengzhao:
To avoid empty search request, set query value to null when the input value for a GoLink component is empty in both RangeFacet and ExactMatchFacet.


cy:
looks good, could you add `cursor: not-allowed` for disabled, and shorten the commit message (one line) See https://github.com/NCI-GDC/portal-ui/blob/next/CONTRIBUTING.md#commits for examples.

cy:
can you rebase/squash and merge, and make sure PRTL-2083 is in the commit message

PR #2328, opened 2018-06-27
 [PRTL-2054] Set max width and overflowTooltip to primary site column

yuanhengzhao:
Added naxWidth with 130 px for primary_site column in ExploreCasesTable and added a tooltip to show the full content if overflow happens.


cy:
can you use this https://github.com/NCI-GDC/portal-ui/blob/next/src/packages/%40ncigdc/uikit/OverflowTooltippedLabel.js `OverflowTooltippedLabel` actually it detects `el.clientWidth < el.scrollWidth` and sets a tooltip

PR #2331, opened 2018-07-09
Added two step warning when download controlled files.

yuanhengzhao:


src/packages/@ncigdc/utils/constants.js
```diff
@@ -222,3 +222,19 @@ export const IMPACT_SHORT_FORMS = {
     unknown: 'UN',
   },
 };
+
+export const FAKE_USER = localStorage.ALLOW_FAKE_USER
```
cy: this doesn't allow real downloads to happen because that requires a token like we discussed but could you load this key from process.env instead of localStorage so user doesn't have a chance to load fake user and scare themselves? :P

src/packages/@ncigdc/components/UserDropdown.js
```diff
@@ -36,16 +36,23 @@ const DropdownItemStyled = styled(DropdownItem, {
 });
 
 const logout = () => {
+  const isDev = process.env.NODE_ENV === 'development';
```
cy: can you set this as an exported constant in constants.js like https://github.com/NCI-GDC/portal-ui/blob/85a91556f92d603072fa5b16f4e71f2173d914c0/src/packages/%40ncigdc/utils/constants.js#L21 then you don't need to keep redefining it from process.env

src/packages/@ncigdc/components/CartDownloadDropdown.js
```diff
@@ -60,31 +60,24 @@ const downloadCart = ({
   setState: Function,
 }) => {
   const { authorized, unauthorized } = authPartitionFiles({ user, files });
-  if (unauthorized.doc_count > 0) {
+  if (!user) {
```
cy: this should still check for `unauthorized.doc_count > 0`, reason being there are `files.access = open` files that anyone even annon can download

src/packages/@ncigdc/utils/constants.js
yuanhengzhao: Np.

src/packages/@ncigdc/dux/auth.js
```diff
@@ -15,8 +15,16 @@ const TOKEN_REQUEST = 'gdc/TOKEN_REQUEST';
 const TOKEN_SUCCESS = 'gdc/TOKEN_SUCCESS';
 const TOKEN_FAILURE = 'gdc/TOKEN_FAILURE';
 const TOKEN_CLEAR = 'gdc/TOKEN_CLEAR';
+const isDev = process.env.NODE_ENV === 'development';
```
cy: use this here https://github.com/NCI-GDC/portal-ui/pull/2331/files#diff-eae22e1881beee47dc829cd9b6228a6eR242

src/packages/@ncigdc/components/UserDropdown.js
```diff
@@ -36,16 +37,22 @@ const DropdownItemStyled = styled(DropdownItem, {
 });
 
 const logout = () => {
+  if (IS_DEV) {
+    localStorage.setItem('ALLOW_FAKE_USER', '');
```
cy: now that it's `process.env.ALLOW_FAKE_USER` this wouldn't work anymore, it's fine no need to logout a fake user in dev; would just remove this block

src/packages/@ncigdc/utils/constants.js
```diff
@@ -222,3 +222,21 @@ export const IMPACT_SHORT_FORMS = {
     unknown: 'UN',
   },
 };
+
+export const FAKE_USER = process.env.ALLOW_FAKE_USER
```
cy: can you prefix `ALLOW_FAKE_USER` with `REACT_APP_` to be consistent with the other process.env vars

PR #2333, opened 2018-07-13
[PRTL-1983] Disable Enter to save set when save button is disabled

yuanhengzhao:
Added a warning for NaN value; disabled enter when NaN value.


yuanhengzhao:
Cool @cy

PR #2349, opened 2018-09-04
[PRTL-2266]Demo feedback from Vincent

yuanhengzhao:


src/packages/@ncigdc/modern_components/CancerDistributionBarChart/CancerDistributionBarChart.js
```diff
@@ -362,93 +329,57 @@ export default compose(
...
-                          }}
-                        >
-                          {cna[f.key] ? '✓' : <span>&nbsp;</span>}
-                        </span>
-                        {f.name}
-                      </label>
-                    ))}
-                </Column>
+              <Column>
+                {/* <CollapsedLegend
```
anncatton: just delete this legend for now since it's not being implemented atm

src/packages/@ncigdc/modern_components/CancerDistributionBarChart/CancerDistributionBarChart.js
```diff
@@ -362,93 +329,57 @@ export default compose(
...
+                          display: 'inline-block',
+                          marginRight: '6px',
+                          marginTop: '3px',
+                          verticalAlign: 'middle',
+                          lineHeight: '16px',
+                        }}
+                      >
+                        {cnv[f.key] ? '✓' : <span>&nbsp;</span>}
+                      </span>
+                      {f.name}&nbsp;&nbsp;&nbsp;
```
anncatton: Add padding for these with css

src/packages/@ncigdc/modern_components/GenesBarChart/GenesBarChart.js
```diff
@@ -183,170 +184,54 @@ const Component = compose(
         onClick: () => handleClickGene(g, mutatedGenesChartData),
       }));
     const checkers = [
```
anncatton: maybe we can extract this hash map out since you're reusing it in the distribution bar chart too?

src/packages/@ncigdc/modern_components/GenesBarChart/GenesBarChart.js
```diff
@@ -402,53 +287,78 @@ const Component = compose(
...
+                            width: '18px',
+                            cursor: 'pointer',
+                            display: 'inline-block',
+                            marginRight: '6px',
+                            marginTop: '3px',
+                            verticalAlign: 'middle',
+                            lineHeight: '16px',
+                          }}
+                        />
+                        {f.name}&nbsp;&nbsp;&nbsp;
```
anncatton: Add padding/styling with css

src/packages/@ncigdc/components/Legends.css
```diff
@@ -0,0 +1,42 @@
+.round {
```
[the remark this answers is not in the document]
yuanhengzhao: It's for the circle legends. Should be removed.

src/packages/@ncigdc/modern_components/GenesBarChart/GenesBarChart.js
yuanhengzhao: Yes. Should do that.

src/packages/@ncigdc/modern_components/GenesBarChart/GenesBarChart.js
```diff
@@ -182,171 +184,51 @@ const Component = compose(
...
-          label: g.symbol,
-          value:
+          symbol: g.symbol,
+          deep_loss:
             context === 'project' && projectId
-              ? score / numCasesAggByProject[projectId] * 100
-              : score / filteredCases.hits.total * 100,
-          tooltip: tooltipContext(context, { symbol: g.symbol, score: score }),
-          onClick: () => handleClickGene(g, cnaGenesChartData),
+              ? g.deep_loss / numCasesAggByProject[projectId] * 100
```
anncatton: Can we pull these calculations into a helper function? Like:
 contextPercentageByCases = (change, aggs) => change/aggs
something like that?

src/packages/@ncigdc/modern_components/CancerDistributionTable/CancerDistributionTable.relay.js
```diff
@@ -17,6 +27,23 @@ export default (Component: ReactClass<*>) =>
               value: 'ssm',
             },
           ]),
+          cnvTested: makeFilter([cnvAvailableVariationDataFilter]),
+          cnvGainFilter: makeFilter([
+            cnvAvailableVariationDataFilter,
+            geneFilter,
```
anncatton: Will need the same fix as in CancerDistributionBarChart!

src/packages/@ncigdc/utils/filters/index.js
```diff
@@ -235,17 +235,20 @@ export const makeFilter: TMakeFilter = fields => {
...
-      return {
-        op: 'in',
-        content: {
-          field: item.field,
-          value,
-        },
-      };
-    }),
+    content: fields
+      .filter(item => item && Object.keys(item).length !== 0)
```
anncatton: I'm not sure this is what should be changed (although having the extra check might not be a bad idea)? there is a filter being passed in that is just an empty object because there's no gene_id on the mutation page, and that's what's breaking here. I was thinking more that the geneFilter you're creating in the cancer distribution queries shouldn't be passed in unless it has values. What do you think?

yuanhengzhao: yes you are right. I will add something like this : if (c.content.field === 'genes.gene_id' && c.content.value)

PR #2357, opened 2018-09-17
[PRTL-2273] Add Legend to svg for both histogram in explore gene table and gene page

yuanhengzhao:


src/packages/@ncigdc/utils/wrapSvg.js
```diff
@@ -177,6 +179,12 @@ export const wrapSvg: TWrapSvg = ({
         )}
       </g>
       ${afterObject.html}
+      <foreignobject  class="node" x="0" y="${sum([
```
anncatton: Can we add this in only if the legends prop is defined?

PR #2370, opened 2018-10-05
[PRTL-2273] Added legends to svg in all barcharts

yuanhengzhao:


src/packages/@ncigdc/utils/wrapSvg.js
```diff
@@ -177,6 +179,12 @@ export const wrapSvg: TWrapSvg = ({
         )}
       </g>
       ${afterObject.html}
+      <foreignobject  class="node" x="0" y="${sum([
```
anncatton: Is there a way to add this only when there is a legend displayed? (when there's a legends prop)

src/packages/@ncigdc/modern_components/GenesBarChart/GenesBarChart.js
```diff
@@ -281,6 +282,31 @@ const Component = compose(
           onClick: () => handleClickGene(g, cnvGenesChartData),
         };
       });
+    const Legends = () => (
```
anncatton: this looks like the same code as above, can you extract it into a component?

PR #2391, opened 2018-10-25
feature/PRTL-2290-fix-githut-labels

anncatton:


src/packages/@ncigdc/components/GitHut/draw.js
```diff
@@ -266,8 +269,8 @@ export default function(params) {
 
     languageLabel
       .append('rect')
-      .attr('x', -padding.left)
-      .attr('width', padding.left)
+      .attr('x', -padding.left - 25)
```
lepsalex: What is 25? It seems to be used in a couple place, consider giving it a const?

PR #2405, opened 2018-11-07
[PRTL-2324] access control banner

yuanhengzhao:


src/packages/@ncigdc/components/Modals/CheckBoxModal.js
```diff
@@ -0,0 +1,83 @@
+// @flow
```
lepsalex: CheckBoxModal.**tsx** ? :nerd_face:

yuanhengzhao: @lepsalex yes. They all should be ts/tsx.

PR #2447, opened 2019-01-17
remove auth_api const, route everything through api. send creds if us…

anncatton:
…er or auth portal


caravinci:
Closed as side effect of branch pruning. Restored and reopened in case this is still relevant.

PR #2485, opened 2019-03-19
Prtl 2482 calculate missing from hits

anncatton:


anncatton:
Also fixes for PRTL-2480, PRTL-2481, PRTL-2486

PR #2497, opened 2019-03-28
Gdcdevops 14/containerize

khan08:
Don't know where this should fit in. Creating a draft for now,


build.sh
```diff
@@ -0,0 +1,15 @@
...
+export REACT_APP_GDC_DISPLAY_SLIDES=true
+export REACT_APP_SLIDE_IMAGE_ENDPOINT="https://api.gdc.cancer.gov/tile/"
+export REACT_APP_COMMIT_HASH=`git rev-parse --short HEAD`
+export REACT_APP_COMMIT_TAG=`git tag -l --points-at HEAD`
+export REACT_APP_API="https://api.gdc.cancer.gov/v0/"
+export REACT_APP_GDC_AUTH="https://portal.gdc.cancer.gov/auth/"
+export GDC_BASE="/"
+export ES_HOST="http://elasticsearch.service.consul:9200"
+export ES_INDEX_VERSION="gdc-r7"
+export REACT_APP_AWG=false
```
singergr: These last two variables should either be omitted from the build script or explicitly unset:
```bash
unset REACT_APP_AWG
unset REACT_APP_IS_AUTH_PORTAL
```
Setting them to `false` will cause them to be parsed as the literal string `"false"`, which will evaluate to true.

Seriously. :(

build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Copied from tungsten
+export REACT_APP_GDC_DISPLAY_SLIDES=true
+export REACT_APP_SLIDE_IMAGE_ENDPOINT="https://api.gdc.cancer.gov/tile/"
+export REACT_APP_COMMIT_HASH=`git rev-parse --short HEAD`
+export REACT_APP_COMMIT_TAG=`git tag -l --points-at HEAD`
+export REACT_APP_API="https://api.gdc.cancer.gov/v0/"
+export REACT_APP_GDC_AUTH="https://portal.gdc.cancer.gov/auth/"
+export GDC_BASE="/"
+export ES_HOST="http://elasticsearch.service.consul:9200"
```
singergr: We should take these `ES_` variables out of here and also out of Tungsten. They aren't actually used.

build.sh
```diff
@@ -0,0 +1,15 @@
...
+export ES_INDEX_VERSION="gdc-r7"
+export REACT_APP_AWG=false
+export REACT_APP_IS_AUTH_PORTAL=false
+npm run build
```
singergr: Before running npm, we should also export these other variables that previously were set in the Travis build:
```bash
NODE_ENV=production
REACT_APP_COMMIT_HASH=$TRAVIS_COMMIT
```

build.sh
ajinkyakadam: this is really sad :(

.travis.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         - npm test || travis_terminate 1
     - stage: deploy
       script:
-        - NODE_ENV=production REACT_APP_COMMIT_HASH=$TRAVIS_COMMIT npm run build
+        - bash build.sh
         - docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" quay.io
```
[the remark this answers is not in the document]
khan08: The encrypted env variables in travis is encrypted per repo's key pair. So we can't copy them from gdcapi. We need to create them in portal repo.

.travis.yml
```diff
@@ -11,17 +11,18 @@ services:
...
-    - stage: deploy
-      script:
-        - NODE_ENV=production REACT_APP_COMMIT_HASH=$TRAVIS_COMMIT npm run build
-        - docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" quay.io
-        - bash docker.sh
+  - stage: test
+    script:
+    # this will only run lint / tests against modified files but it is unstable
+    # - echo $TRAVIS_COMMIT_RANGE | sed 's/\.\.\./../' | xargs git diff-tree --no-commit-id --name-only -r | xargs echo -n
+    - npm run lint -- --max-warnings=0 || travis_terminate 1
```
singergr: Looks like something changed with the build and the `--max-warnings` option isn't accepted anymore. Probably better to update this part to match what's in develop now (I think this option was there way back when the PR was first opened and then removed later).

Dockerfile
```diff
@@ -2,6 +2,4 @@ FROM nginx:stable-alpine
 
 RUN rm -v /etc/nginx/conf.d/default.conf
 
-COPY infra/nginx.conf /etc/nginx/conf.d/nginx.conf
-
 COPY build /usr/share/nginx/html
```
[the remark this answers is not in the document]
singergr: I'd also suggest using the latest release tag rather than `latest`. Currently, that'd be `1.10.3-oobleck`.

Dockerfile
```diff
@@ -1,7 +1,3 @@
-FROM nginx:stable-alpine
-
-RUN rm -v /etc/nginx/conf.d/default.conf
```
singergr: We'll overwrite this file when we run the container, so taking out this `rm` is fine. However, we should also remove `/etc/nginx/sites-enabled/default` so that it won't set up any locations that we don't need.

PR #2501, opened 2019-04-08
[PRTL-2506] Fix table height on scroll

yuanhengzhao:


src/packages/@ncigdc/containers/explore/ClinicalAggregations.tsx
```diff
@@ -280,7 +280,7 @@ const enhance = compose(
       dispatch,
       notifications,
     }: IClinicalProps): any => {
-      return [
+       return (<Column>
```
caravinci: Maybe this could be a Fragment instead? That way we don't add more HTML elements/markup.

PR #2510, opened 2019-04-15
PRTL-2517 remove unnecessary layout breaking scrollbars

caravinci:
Leaving it at `overflow: auto` seems to take care of the problem well enough.
Will amend if testing finds unforeseen side-effects.


caravinci:
which means the change doesn't seem to break experience on a mac as a side effect. awesome

PR #2516, opened 2019-04-16
[PRTL-2459] Update References to Moved Properties for Pinwheel

yuanhengzhao:


anncatton:
We'll merge this in once we've tested the data in pinwheel

PR #2517, opened 2019-04-17
[PRTL-2503] Print/download Clinical Analysis plots

samrichca:


src/packages/@ncigdc/theme/icons/Print.js
```diff
@@ -0,0 +1,6 @@
+// @flow
```
anncatton: Can get rid of flow here.

PR #2519, opened 2019-04-22
PRTL-2529 fix user profile crash when no projects are assigned

caravinci:


caravinci:
Just noticed a few long deprecated lifecycles break with the React upgrade.
Will go and fix them but this cannot be merged until that's tested more thoroughly.

caravinci:
Nope. Removed yarn and the react upgrades (which would've fixed the devtools naming issues). Too large a scope for this task; thus bugfix is complete as per ticket spec.

PR #2521, opened 2019-04-24
PRTL-2500 add “no data” component to empty categories

caravinci:
WIP. this component can and should be abstracted into a reusable component.


src/packages/@ncigdc/containers/explore/ClinicalAggregations.tsx
```diff
@@ -443,82 +456,93 @@ const enhance = compose(
                     </Row>
                     {facetsExpandedStatus[facet.field].expanded && (
                       <Column>
-                        {_.orderBy(
```
caravinci: this is where the actual magic happens 😆

PR #2528, opened 2019-04-30
[PRTL-1896] rotate & truncate histogram X-axis labels

samrichca:
for review: look at the 'rotate & truncate' commit, and this

https://github.com/NCI-GDC/portal-ui/pull/2528/commits/4015fc708c3123c7cdcb99529b4479557e3652d3#diff-02b7382bb8999d242d665f1fd7442f99L95

### DIFF
--- a/src/packages/@ncigdc/components/Charts/BarChart.js
+++ b/src/packages/@ncigdc/components/Charts/BarChart.js
@@ -8,8 +8,6 @@ import {
   compose,
   withState,
   pure,
-  withProps,
-  withPropsOnChange,
 } from 'recompose';
 
 // Custom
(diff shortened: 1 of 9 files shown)


src/packages/@ncigdc/components/Charts/BarChart.js
```diff
@@ -18,6 +16,8 @@ import { withTooltip } from '@ncigdc/uikit/Tooltip';
 import withSize from '@ncigdc/utils/withSize';
 import './style.css';
 
+export const DEFAULT_X_AXIS_LENGTH = 8;
```
anncatton: thank you for naming this! can we add `LABEL` to the name? otherwise sounds like the axis range :)

PR #2538, opened 2019-05-10
release/1.21.0 into develop.

yuanhengzhao:


package.json
```diff
@@ -119,4 +119,4 @@
     "ts-loader": "5.2.2",
     "typescript": "3.1.3"
   }
-}
+}
```
caravinci: ?

anncatton: hmm i don't know why my linter would have done that 🤔

PR #2545, opened 2019-05-15
fix api url

anncatton:


build.sh
```diff
@@ -1,15 +1,14 @@
 #!/bin/bash
 # Copied from tungsten
 export REACT_APP_GDC_DISPLAY_SLIDES=true
-export REACT_APP_SLIDE_IMAGE_ENDPOINT="https://api.gdc.cancer.gov/tile/"
+export REACT_APP_SLIDE_IMAGE_ENDPOINT="https://portal.gdc.cancer.gov/auth/api/tile/"
```
singergr: We're actually still pointing this one directly to the API. Probably better to leave it like that.

anncatton: Gotcha. I'll revert.

PR #2575, opened 2019-06-10
[PRTL-2566] Added Range to Continuous Binning Query

yuanhengzhao:
## Ticket Number

e.g. PRTL-2566

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
Added Range to Continuous Binning Query

### DIFF
--- a/src/packages/@ncigdc/modern_components/ClinicalAnalysis/ContinuousAggregationQuery.js
+++ b/src/packages/@ncigdc/modern_components/ClinicalAnalysis/ContinuousAggregationQuery.js
@@ -28,8 +28,8 @@ const getContinuousAggs = ({ fieldName, stats, filters, bins }) => {
 
   let rangeArr = _.reduce(bins, (acc, bin, key) => {
     if (
-      bin.from &&
-      bin.to &&
+      (typeof bin.from === 'number') &&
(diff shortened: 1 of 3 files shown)


src/packages/@ncigdc/modern_components/ClinicalAnalysis/ContinuousAggregationQuery.js
```diff
@@ -20,22 +20,52 @@ const simpleAggCache = {};
...
+const getContinuousAggs = ({ fieldName, stats, filters, bins }) => {
   // prevent query failing if interval will equal 0
   if (_.isNull(stats.min) || _.isNull(stats.max)) {
     return null;
   }
-  const interval = (stats.max - stats.min) / DEFAULT_CONTINUOUS_BUCKETS;
+
+  let rangeArr = _.reduce(bins, (acc, bin, key) => {
+    if (
+      bin.from &&
```
samrichca: `from` or `to` could be 0 which would make this condition fail

yuanhengzhao: Just find out you put comment here. Yes. You are right. I will correct it.

PR #2580, opened 2019-06-18
Cohort builder into UAT (develop)

anncatton:
## Ticket Number

UAT test

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
UAT testing


.prettierrc.yaml
```diff
@@ -0,0 +1,4 @@
+trailingComma: 'all'
+tabWidth: 2
+semi: true
+singleQuote: true
```
caravinci: sorry if it is a silly question, but why is this file here? 
I thought we had removed prettier altogether...

PR #2617, opened 2019-07-11
[PRTL-2681] only validate the last row if it has input

samrichca:
## Ticket Number

PRTL-2681

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

only validate the last row if it has input


src/packages/@ncigdc/components/Modals/ContinuousBinning/RangeInputRow.js
```diff
@@ -192,17 +204,19 @@ class RangeInputRow extends React.Component {
...
+    const allFieldsEmpty = checkEmptyFields(fieldValues);
+
     return (
       <OutsideClickHandler
         disabled={!rangeMethodActive}
         onOutsideClick={() => {
           this.handleValidate();
         }}
-        >
+      >
```
caravinci: Can we fix this on VSCode? Henry's commits do the same.

PR #2632, opened 2019-07-18
🐛 PRTL-2546 add consequence_type fallback to survival plot slug

anncatton:
## Ticket Number

PRTL-2546

- [x] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
Survival plot slug will display `consequence_type` when `aa_change` is not present

### DIFF
--- a/src/packages/@ncigdc/modern_components/SsmsTable/SsmsTable.model.js
+++ b/src/packages/@ncigdc/modern_components/SsmsTable/SsmsTable.model.js
@@ -1,4 +1,5 @@
 // @flow
+/* eslint-disable react/display-name */
 
 import React from 'react';
 import { scaleOrdinal, schemeCategory10 } from 'd3';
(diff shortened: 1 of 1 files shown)


src/packages/@ncigdc/modern_components/SsmsTable/SsmsTable.model.js
```diff
@@ -300,8 +332,8 @@ const SsmsTableModel = [
                     'consequence.hits.edges[0].node.transcript.gene.symbol'
                   )} ${get(
```
anncatton: this is the actual change, everything else is formatting

caravinci: It feels `get` should let us pass it alternate variables to be found, rather than having to 'get' multiple times... right?

anncatton: yeah, it takes a third arg for a default, but doesn't allow that to be another `get`...unless there is another way of using this i'm missing

PR #2638, opened 2019-07-23
💄 PRTL-2625 styling updates for bin modals. Set default continuous va…

anncatton:
…lue props for onMount

## Ticket Number

PRTL-2625

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
Styling updates for continuous and categorical modals
Sets default prop values in continuous modal on mount to prevent crash (occurs with older settings in analysis redux)
Text changes

### DIFF
--- a/src/packages/@ncigdc/components/Modals/ContinuousBinning/ContinuousCustomBinsModal.js
+++ b/src/packages/@ncigdc/components/Modals/ContinuousBinning/ContinuousCustomBinsModal.js
@@ -58,7 +58,7 @@ class ContinuousCustomBinsModal extends Component {
   componentDidMount = () => {
     const {
       continuousBinType = 'default',
-      continuousCustomInterval = 0,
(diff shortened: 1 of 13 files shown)


src/packages/@ncigdc/components/Modals/ContinuousBinning/ContinuousCustomBinsModal.js
```diff
@@ -55,9 +57,9 @@ class ContinuousCustomBinsModal extends Component {
 
   componentDidMount = () => {
     const {
-      continuousBinType,
-      continuousCustomInterval,
-      continuousCustomRanges,
+      continuousBinType = 'default',
+      continuousCustomInterval = 0,
```
samrichca: Custom interval is an object now

{ Amount, max, min }

So it can be preserved in Redux when you switch to custom range bins

PR #2691, opened 2019-09-04
[PRTL-2746] fix print issues

samrichca:
## Ticket Number

PRTL-2746

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

- add print CSS rules for the box plot
- make SVGs exempt from print CSS colour rules - print them as designed


src/packages/@ncigdc/modern_components/ClinicalAnalysis/ClinicalVariableCard/ClinicalVariableCard.js
```diff
@@ -315,7 +314,7 @@ const ClinicalVariableCard = ({
                 justifyContent: 'center',
               }}
               >
-              There is no data for this facet
+              <div className="print-mb print-mt">There is no data for this facet</div>
```
caravinci: Could a `span` work, instead of `div` (considering the text is meant as an inline element)?

samrichca: I had to use a block level element so I could put margins on the element in print CSS

caravinci: 🤔 that makes sense...
But in that case, shouldn't the classes simply be applied to the parent, rather than nesting another div inside it? (not sure what i'm saying would work, but would avoid adding more `div`s)

samrichca: the parent is a pre-made `Row` component and I want to avoid interactions with other classes

PR #2695, opened 2019-09-05
[PRTL-2742] Analysis Layout Enhancements & Fixes

yuanhengzhao:
## Ticket Number

PRTL-2742

- [ ] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
1. Abandon ClinicalAnalysisLaunch.tsx file and move all its union logic into SelectedSet.js file since they have 80% common code. 
2. Correct duplicate variables errors in SelectedSet.js file.
3. Apply layout style of ClinicalAnalysisLaunch to selectedSet.
4. Add textOverflow: 'ellipsis' and tooltips to CohortDropdown.
5. Fix a checkbox error which is for selecting sets component, when uncheck a checkbox, all the selected checkbox will be unselected.

### DIFF
--- a/src/packages/@ncigdc/components/analysis/CreateAnalysis.js
+++ b/src/packages/@ncigdc/components/analysis/CreateAnalysis.js
@@ -6,17 +6,13 @@ import {
   withState,
 } from 'recompose';
 import { connect } from 'react-redux';
-
 import { addAnalysis } from '@ncigdc/dux/analysis';
 import withRouter from '@ncigdc/utils/withRouter';
 import Button from '@ncigdc/uikit/Button';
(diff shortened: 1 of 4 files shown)


src/packages/@ncigdc/components/analysis/CreateAnalysis.js
```diff
@@ -89,10 +81,10 @@ const CreateAnalysis = ({
           padding: '2rem 2.5rem',
         }}
         >
-        {availableAnalysis.map(analysis => {
+        {availableAnalysis.map(a => {
```
caravinci: Declarative variable names are more desirable in this case. Can we keep those?

PR #2709, opened 2019-09-12
[PRTL-2757] Add Summary Tab Layout Wrapper

yuanhengzhao:
## Ticket Number

PRTL-2757

- [ ] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
Add title and download button to the layout Wrapper, which is named CardWrapper. Apply tooltips in this Wrapper so that all card's tooltips are controlled by this wrapper.


src/packages/@ncigdc/components/Explore/SummaryPage/SampleTypeCard.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PieChart from '@ncigdc/components/Charts/PieChart';
+import * as d3 from 'd3';
+
+const color = d3.scaleOrdinal(d3.schemeCategory20);
+
+const SampleTypeCard = ({ data, mappingId }) => (
+  // <div
```
caravinci: no need to leave this commented-out code

src/packages/@ncigdc/components/Explore/SummaryPage/HistogramCard.js
```diff
@@ -0,0 +1,62 @@
...
+const CHART_HEIGHT = 320;
+
+const HistogramCard = ({
+  bottomMarginForxAxisTitle,
+  data,
+  mappingLabel,
+  mappingValue,
+  xAxisTitle,
+}) => {
+  const CHART_MARGINS = {
```
caravinci: this can be extracted from the function (i.e. before line 22), so a new `const` is not created at every render.

PR #2714, opened 2019-09-17
[PRTL-2742]Analysis Layout Enhancements & Fixes

yuanhengzhao:
## Ticket Number

e.g. PRTL-0000

- [ ] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)


src/packages/@ncigdc/modern_components/ClinicalAnalysis/CohortDropdown.js
```diff
@@ -24,22 +24,34 @@ const CohortDropdown = ({
...
-          dispatch(
-            updateClinicalAnalysisSet({
-              id: currentAnalysis.id,
-              setId: setKey,
-              setName: name,
-            })
-          );
-        }}
+      <Tooltip
+        Component={name.length > 24 ? name : null}
```
anncatton: i think we need to reduce this length check to about 20 or 21, the tooltip doesn't show on some names that have the ellipsis overflow

PR #2719, opened 2019-09-23
Summary page feature

anncatton:
## Ticket Number

n/a

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
Adds summary page to Explore, hide behind feature flag


src/packages/@ncigdc/components/Layouts/MasonryLayout.tsx
```diff
@@ -17,10 +17,7 @@ interface IMasonryLayoutProps {
 
 
 const MasonryLayout = ({
-  customizeStyle = {
-    borderStyle: 'solid', // Replace it with your own style
-    borderWidth: '1px',
-  },
+  customizeStyle, // Replace it with your own style
```
anncatton: can we add a default `{}` back in?

src/packages/@ncigdc/components/SearchPage.js
```diff
@@ -100,7 +100,7 @@ const SearchPage = (
         )}
         <CurrentFilters style={{ flex: 1 }} {...filtersLinkProps} />
       </Row>
-      {results}
+      {typeof results === 'function' ? results({ showFacets }) : results}
```
anncatton: i think this might have reverted back to pre-`1.23.0` code, `release` branch just has `{results}`

yuanhengzhao: I use this function because in summary page, I need this `showFacets` value to determine whether to show 3 or 4 cards in a row. I can do it in redux too. But it will be more complicated.

anncatton: ok!

src/packages/@ncigdc/components/Layouts/MasonryLayout.tsx
yuanhengzhao: Will fix it.

PR #2737, opened 2019-10-04
[PRTL-2875] improve case guarding for OncoGrid buildOcurrences

caravinci:

- [ ] `prod`
- [x] `dev-oicr`
- [x] `qa` (which version?)

## Description of Changes
Another case of mismatch between the unnormalised genes vs the normalised data.
More case guards. This one was harder to find. Funny how much difference a couple lines can make.


src/packages/@ncigdc/components/Oncogrid/dataMapping.js
```diff
@@ -213,7 +213,7 @@ export const buildOccurrences: TBuildOccurrences = (
   const geneIdToSymbol = {};
   for (let i = 0; i < genes.length; i += 1) {
```
caravinci: I reaaaaaally had to fight refactoring this into an Array.reduce. ugh

PR #2753, opened 2019-10-29
[PRTL-2887] add number check to print p-value

samrichca:
## Ticket Number

e.g. PRTL-0000

- [x] `prod`
- [x] `dev-oicr`
- [x] `qa-yellow`

## Description of Changes

- add default for `buttonStyle` to fix a page crash
- change `pValue` check - look for a number, not just a truthy value
- refactor - import specific lodash functions


src/packages/@ncigdc/components/SurvivalPlotWrapper.js
```diff
@@ -211,7 +211,6 @@ const SurvivalPlotWrapper = ({
                 ))}
             </Row>
           </div>
-
```
caravinci: Why is this change necessary?

I was trying to fix it using github, but it messes the spacing.
Mind undoing it?

samrichca: merge funkiness probably. hang on

PR #2766, opened 2019-11-18
[PRTL-2915] stop continuous card re-mounting

samrichca:
## Ticket Number

PRTL-2915

- [x] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

- add `key` prop to continuous and categorical card, because they're list items and changing the list is causing them to re-mount
- make sure overall survival only updates when necessary


src/packages/@ncigdc/modern_components/ClinicalAnalysis/ClinicalAnalysisResult.js
```diff
@@ -433,7 +433,8 @@ export default compose(
       overallSurvivalData,
       survivalDataLoading,
     }) => (
-      !isEqual(currentAnalysis, nextCurrentAnalysis) ||
```
samrichca: this change causes overall survival to update ONLY if you switch to a new analysis or case set.

previously, it would update if you made custom bins in the gender card, even though that has nothing to do with overall survival.

src/packages/@ncigdc/modern_components/ClinicalAnalysis/ClinicalAnalysisResult.js
```diff
@@ -488,7 +491,8 @@ export default compose(
         nextControlPanelExpanded === controlPanelExpanded &&
         nextLoading === loading &&
         isEqual(populateSurvivalData, nextPopulateSurvivalData) &&
-        nextSurvivalDataLoading === survivalDataLoading
```
samrichca: added diff check for `currentAnalysis` here, because `overallSurvival` is no longer listening to changes to `currentAnalysis` - without this, the page won't update when you make custom bins or switch to a different case set.

src/packages/@ncigdc/modern_components/ClinicalAnalysis/ClinicalVariableCard/ClinicalVariableCard.js
```diff
@@ -242,6 +243,7 @@ const ClinicalVariableCard = ({
         padding: '0.5rem 1rem 1rem',
         ...style,
       }}
+      key={key}
```
samrichca: this is what stops the card from re-mounting - cards are in a list, they need keys

src/packages/@ncigdc/modern_components/ClinicalAnalysis/ClinicalVariableCard/ContinuousVariableCard.js
```diff
@@ -279,7 +279,6 @@ export default compose(
           axisTitle: dataDimensions[sanitisedId].axisTitle,
           boxPlotValues: map(
             {
-
               ...rawQueryData.stats,
```
samrichca: fixing weird formatting

src/packages/@ncigdc/modern_components/ClinicalAnalysis/ContinuousAggregationQuery.js
```diff
@@ -255,11 +255,12 @@ export default compose(
     }
   ),
 )(({
-  aggData, hits, isLoading, setId, stats, ...props
+  aggData, hits, isLoading, key, setId, stats, ...props
```
samrichca: this is going to get removed in the PR for PRTL-2917

samrichca: this was updated in "Merge branch 'feat/continuous-card-refactor' into prtl-2915"

samrichca: resolved by merging PRTL-2917

src/packages/@ncigdc/modern_components/ClinicalAnalysis/ClinicalVariableCard/ClinicalVariableCard.js
anncatton: i don't think the `key` is getting passed all the way down from the card components in `ClinicalAnalysisResult` - i'm seeing `undefined` when i log it and the cards are still remounting.

src/packages/@ncigdc/modern_components/ClinicalAnalysis/ClinicalVariableCard/ClinicalVariableCard.js
```diff
@@ -167,6 +167,7 @@ const ClinicalVariableCard = ({
   filters,
   id,
   isLoading,
+  key,
```
anncatton: i think you can remove this one too, it's not being used

PR #2777, opened 2019-12-04
[PRTL-2753] make header calculate its own height dynamically

caravinci:

- [x] `prod`
- [x] `dev-oicr`
- [x] `qa` (any version)

## Description of Changes
Duplicates the header invisibly (and disabling keyboard access), while keeping a relative positioning; which ensures the content will be pushed up/down when the header changes height.

Not the most elegant solution, but programmatic height detection and updating would require considering far too many possible interactions, leading to possible -unnecessary- performance issues.


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.23.0",
+  "version": "1.23.1",
```
caravinci: we forgot this one with the hotfix :)

src/packages/@ncigdc/components/PortalContainer.js
```diff
@@ -62,10 +57,8 @@ const PortalContainer = ({
       id="skip"
       role="main"
       style={{
+        minWidth: 1024,
         paddingBottom: '120px',
-        paddingTop: `calc(51px + ${notifications.filter(n => !n.dismissed)
```
caravinci: Instead of having the padding be calculated manually with a fixed multiplier, we let the browser carry the weight using css flow.

src/packages/@ncigdc/components/PortalContainer.js
```diff
@@ -62,10 +57,8 @@ const PortalContainer = ({
       id="skip"
       role="main"
       style={{
+        minWidth: 1024,
```
caravinci: Relocated this line from 54 to sure the minimum weight is *only* on the container for the elements that currently care for the website not being responsive.

caravinci:
> are the header & spacer header identical other than `visibility: hidden;`? if so can they be made into a reusable component?

Agreed 100%. Alas, keeping scope.

caravinci:
> are the header & spacer header identical other than `visibility: hidden;`? if so can they be made into a reusable component?

Agreed, and even the AWG header can be merged into the regular header, with some flag modifiers. 

Pretty much, expect the spacer doesn't have any of the click/tab functionalities either (i.e. just the layout + content); to avoid issues with alternate input methods.

Tried to keep the scope to resolving the observed issue, while leaving refactoring or improvement to a later stage of the app.

samrichca:
keeping our code clear and dry is in-scope. I would suggest keeping AWG and portal separate per @anncatton 's advice, and create a reusable header for each.

my concern is the difference between the duplicate headers won't be obvious to developers working on it in the future, creating the potential for bugs. the header is a critical, universal component and there's no rush.

caravinci:
> keeping our code clear and dry is in-scope. 

> the header is a critical, universal component and there's no rush.

Fair enough. That is 100% a good argument. Will work on refactoring this into a cleaner solution.

-------

> I would suggest keeping AWG and portal separate per @anncatton 's advice, and create a reusable header for each.

> my concern is the difference between the duplicate headers won't be obvious to developers working on it in the future, creating the potential for bugs. 

I'm afraid I've missed where/when that advice was given, yet seems to me the only difference between AWGHeader and the plain regular header, is that the latter has added features/code; otherwise both are functional copies of each other. 
i.e. the difference between AWGHeader and Regular is not that obvious, and could instead be handled by using flags in a single component, rather than duplicating code with bits and pieces that somehow differentiate them, as you **correctly** suggest should be done in this PR.

However, as per Header vs AWGHeader, the duplication added in this PR is merely following the app's already established pattern. 

--------

Blocking this PR while amending this code to follow the suggestion. Thanks! 👍

caravinci:
Alright... I've updated the PR by removing the duplicate elements, and instead used a ResizeObserver polyfill (yes, safari + ie + edge) to leverage header size changes as an event; and used that new height to grow an empty placeholder `div`. 
TL;DR: No duplication, and the polyfill keeps it pretty performant too.

PR #2825, opened 2020-03-05
[PRTL-3039] add table to controlled access modal

samrichca:
Merge PRTL-3041 first!

## Ticket Number

PRTL-3039

- [x] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

- add a table that shows which controlled programs are available
- add buttons so we can switch between scenarios (logged in, logged out, some controlled access, etc) while we don't have API data. I made this a component because this is a recurring need.


src/packages/@ncigdc/components/LoginButton.js
```diff
@@ -19,13 +19,21 @@ const Link = styled.a({
   transition: 'background-color 0.2s ease',
 });
 
-const LoginButton = ({ children, dispatch, user }) => (
+const styles = {
```
caravinci: is this being used?

samrichca: nope, hang on

src/packages/@ncigdc/components/LoginButton.js
```diff
@@ -19,13 +19,21 @@ const Link = styled.a({
...
+  marginLeft: {
+    marginLeft: '0.7rem',
+  },
+};
+
```
samrichca:
```suggestion
```

PR #2838, opened 2020-04-01
[PRTL-3049] integrate controlled access components

caravinci:

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
- Creates a new "withControlledAccess" HOC, which isolates all the controller logic for DAVE-CA functionality.
- Modifies relevant parts of the exploration section to receive props from said HOC, to feed the newly added components (CA switch, section banner/title).
- Abstracts the modal as a direct dependency of that HOC, rather than a direct injection into exploration, which allows for it to be summoned from anywhere in the portal.

Note: It removes the modal window selection behaviour (i.e. when you press the "explore" button and the modal closes), as it will be addressed in PRTL-3050, along with the corresponding URL query parametrisation.


src/packages/@ncigdc/components/Modals/ControlledAccess/ControlledAccessModal.js
```diff
@@ -1,134 +1,111 @@
...
+    user: state.auth.user,
+  })),
+  withState(
+    'selectedStudies',
+    'setSelectedStudies',
+    ({
+      activeControlledPrograms,
+      userAccessList,
+    }) => (userAccessList.length === 1 ? userAccessList : activeControlledPrograms),
+    // TODO: this ^^^ is a placeholder for the "controlled" array of the CA response.
```
caravinci, line 97: @samrichca Having been sure not to remove logic that was there, I've gone looking deeper, and remembered I'd already added this placeholder foreseeing the stuff to be handled later.

PR #2849, opened 2020-05-11
[PRTL-3055] enable program filter based on CA

samrichca:
## Ticket Number

PRTL-3055

- [ ] `prod`
- [x] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

Enable filter based on CA so by default you only see data for the program you selected. Cosmic filter is suppressed in order to show FM data as that's currently the only controlled program, and Cosmic + FM = no data. 

Also added white background to Explore nav button & removed problematic borderStyle prop from Tab.js.


src/packages/@ncigdc/components/Header/styles.scss
```diff
@@ -142,6 +142,7 @@
     > li > button,
     .header-link {
       align-items: center;
+      background-color: #fff;
```
caravinci, line 145: Strange, I've already added this change, why is Github not picking that up? 🤔

image

src/packages/@ncigdc/components/Modals/ControlledAccess/ControlledAccessModal.js
```diff
@@ -58,45 +73,45 @@ const ControlledAccessModal = ({
...
+      )}
+      title="Explore Controlled & Open Data"
+      >
+      <CAMessage user={user} userAccessList={userAccessList} />
+      <EntityPageHorizontalTable
+        data={formatData({
+          handleProgramSelect,
+          selectedStudies,
+          setSelectedStudies,
+          studiesList,
```
caravinci, line 85: LOL I've missed this cheeky little one. 🙃

PR #2867, opened 2020-05-26
resolve collapsed banner contrast issues

caravinci:
#2248  Environment to be used in testing

- [x] `prod`
- [x] `dev-oicr`
- [x] `qa` (any version)

## Description of Changes
Collapses the text elements inside the collapsed banner, so there's no false positives for contrast issues.

### DIFF
--- a/src/packages/@ncigdc/uikit/Banner/index.js
+++ b/src/packages/@ncigdc/uikit/Banner/index.js
@@ -65,15 +65,16 @@ const Banner = ({
     )}
 
     {dismissible && (
-      <span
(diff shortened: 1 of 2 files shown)


src/packages/@ncigdc/uikit/Banner/index.js
```diff
@@ -63,6 +66,7 @@ const Banner = ({
 
     {dismissible && (
       <span
```
samrichca, line 68: pls make this a button and we're ready to roll

PR #2874, opened 2020-06-10
🐡 Very Bad Fish 

caravinci:


.env
```diff
@@ -8,5 +8,5 @@ REACT_APP_GDC_AUTH=/auth/
 REACT_APP_COMMIT_HASH=
 REACT_APP_COMMIT_TAG=
 DANGEROUSLY_DISABLE_HOST_CHECK=true
-REACT_APP_ALLOW_FAKE_USER=
+REACT_APP_ALLOW_DEV_USER=
```
caravinci, line 11: this is just a rename and it's dev-only so no need to worry, but the app's functionality allows for either name (i.e. legacy compatible)

src/packages/@ncigdc/utils/constants.js
```diff
@@ -343,8 +343,12 @@ export const IMPACT_SHORT_FORMS = {
   },
 };
 
-export const FAKE_USER =
-  localStorage.REACT_APP_ALLOW_FAKE_USER || process.env.REACT_APP_ALLOW_FAKE_USER
+export const IS_DEV = process.env.NODE_ENV === 'development';
+
+export const DEV_USER =
```
caravinci, line 348: dev user renaming is legacy compatible (aka "fake"), as seen here

PR #2876, opened 2020-06-17
[PRTL-3151] remove stray CA request.

caravinci:

- [x] `prod`
- [x] `dev-oicr`
- [x] `qa-orange`

## Description of Changes
- Removes an unnecessary call to one of the CA endpoints, by putting it behind the corresponding feature flag.
Bonus: a couple other fixes to ensure no crashes happen if the flag is added while the CA endpoints are not available.

### DIFF
--- a/src/packages/@ncigdc/components/Modals/ControlledAccess/ControlledAccessModal.js
+++ b/src/packages/@ncigdc/components/Modals/ControlledAccess/ControlledAccessModal.js
@@ -82,7 +82,7 @@ const ControlledAccessModal = ({
           handleProgramSelect,
           selectedStudies,
           setSelectedStudies,
-          studiesList,
+          studiesSummary,
           user,
           userAccessList,
         })}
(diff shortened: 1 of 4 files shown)


src/packages/@ncigdc/components/Modals/ControlledAccess/ControlledAccessModal.js
```diff
@@ -82,7 +82,7 @@ const ControlledAccessModal = ({
           handleProgramSelect,
           selectedStudies,
           setSelectedStudies,
-          studiesList,
+          studiesSummary,
```
caravinci, line 85: This one got renamed and the change was applied to the feature branch, but not onto mainstream.
Modal crashes if the flag is on.

src/packages/@ncigdc/utils/withControlledAccess/index.js
```diff
@@ -70,6 +74,15 @@ export default compose(
       });
     },
   }),
+  lifecycle({
+    componentDidMount() {
+      const {
+        fetchStudiesList,
+      } = this.props;
+
+      DISPLAY_DAVE_CA && fetchStudiesList();
```
caravinci, line 83: This is where the stray call is restrained by the flag.

src/packages/@ncigdc/utils/withControlledAccess/index.js
```diff
@@ -173,29 +186,22 @@ export default compose(
...
       };
     },
   ),
-  lifecycle({
-    componentDidMount() {
-      const {
-        fetchStudiesList,
-      } = this.props;
-
-      fetchStudiesList();
```
caravinci, line 198: The flag was missing here.

PR #2883, opened 2020-07-15
[PRTL-3165] use real API shape in gene expression

samrichca:
## Ticket Number

PRTL-3165

- [x] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

- replace our stub JSON data with the stub data provided by UC
- update naming/formatting to match
- massage data to separate case IDs from categories


src/packages/@ncigdc/modern_components/GeneExpression/GeneExpressionChart.js
```diff
@@ -41,6 +42,7 @@ class GeneExpressionChart extends Component {
     };
     // this doesn't work if jquery is imported
     // in this file. ignore the eslint error
+    // about the $ symbol on the line below
```
caravinci, line 45: this can be avoided adding the global variable to the eslint config, like so:
image

samrichca: nice, thanks!

PR #2893, opened 2020-08-05
GDCDEVOPS-535: Using latest nginx-extras image

a-collet:
## Ticket Number

GDCDEVOPS-535

- [ ] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes
- Updating dockerfile to use latest nginx-extras container


a-collet:
Closing as we're currently running into an issue with helm templates and the length of the branch name

PR #2896, opened 2020-08-14
[PRTL-3190] add demo button for scrna-seq

samrichca:
## Ticket Number

PRTL-3190

- [x] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

add demo button, which is a select button with demo text in this case


caravinci:
Wouldn't it be easier then to use the DemoButton, which already includes the tooltip?

samrichca:
> Wouldn't it be easier then to use the DemoButton, which already includes the tooltip?

@caravinci I thought about it... at the moment scrnaseq only has 5 cases, so we're showing the analysis selection as a demo. haven't been asked to do that before & it'll be changed to the default analysis setup soon. based on that I didn't want to make additional changes to the components/logic for selecting analyses or viewing a demo.

caravinci:
Makes sense. Thanks for clarifying for me.

PR #2904, opened 2020-08-24
[PRTL-3192] add real scrna seq filters

samrichca:
## Ticket Number

PRTL-3192

- [ ] `prod`
- [ ] `dev-oicr`
- [x] `qa-yellow`

## Description of Changes

- add real scrna-seq filters for case/workflow selection


src/packages/@ncigdc/components/analysis/SelectScrnaSeq/SelectScrnaSeq.relay.js
```diff
@@ -4,33 +4,45 @@ import { compose, pure, setDisplayName } from 'recompose';
...
+      content: {
+        field: "files.data_type",
+        value: [
+          "Single Cell Analysis"
         ]
       }
     }
-  ]
+  ],
+  op: "and",
```
caravinci, line 38: Totally nitpicking: any particular reason why these are double quotes?

PR #2911, opened 2020-08-31
[PRTL-3192] use real scrna filters

samrichca:
## Ticket Number

PRTL-3192

- [x] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

use real scRNA-seq filters

the "fake" ones are commented out so we can switch between them


samrichca:
redundant?

PR #2923, opened 2020-09-21
[PRTL-3195] add AWG token refreshing capabilities

caravinci:

- [ ] `prod`
- [x] `dev-oicr`
- [x] `qa-purple`

## Description of Changes
Renews the token automatically upon user interactions, as long as the user is logged in, and at least 5 minutes have passed. It does so by setting a timestamp in local storage, which is updated (or removed) whenever a renewal is requested.
This commit also ensures no duplicate renewal queries are made by throttling simultaneous requests, and caching the latest response for 5 seconds (in case near simultaneous requests are made, and make it through the throttling routine.)

The minimum time to make a request is configurable through an environment flag (default 5 minutes), currently commented out, yet left for reference.


awgbuild.sh
```diff
@@ -5,10 +5,10 @@ export REACT_APP_SLIDE_IMAGE_ENDPOINT="/auth/api/v0/tile/"
 export REACT_APP_COMMIT_HASH=`git rev-parse --short HEAD`
 export REACT_APP_COMMIT_TAG=`git tag -l --points-at HEAD`
 export REACT_APP_API="https://portal.awg.gdc.cancer.gov/auth/api"
+export REACT_APP_FENCE="https://portal.awg.gdc.cancer.gov/fence/"
 export REACT_APP_GDC_AUTH="https://portal.awg.gdc.cancer.gov/auth/"
-export REACT_APP_FENCE="https://login.awg.gdc.cancer.gov"
+# export AWG_TOKEN_EXPIRY=5
```
caravinci, line 10: This flag sets a custom number of minutes the UI should wait before allowing another query to renew the token.
(Default is 5, hardcoded)

src/Root.js
```diff
@@ -141,11 +69,9 @@ window.store = store;
 
 store.dispatch(fetchApiVersionInfo());
 
-if (process.env.NODE_ENV !== 'development') {
+if (!(IS_DEV || AWG)) {
```
caravinci, line 72: None of these actions should be dispatched in AWG on load (before login).

src/Root.js
```diff
@@ -46,91 +59,6 @@ const AccessError = message => {
...
-        window.forceRelayRetry = cb;
-        console.log(
-          `call \`forceRelayRetry()\` for immediately retry! Or wait ${delay} ms.`
-        );
-      },
-      retryDelays: attempt => ((attempt + 4) ** 2) * 100,
-      statusCodes: retryStatusCodes,
-    }),
-    // Add hash id to request
-    next => req => {
```
caravinci, line 66: Abstracting this into a recompose handler, to be passed into the Relay network layer after component mount.

src/Root.js
```diff
@@ -162,58 +88,169 @@ const HasUser = connect(state => state.auth)(props => {
...
+  }),
+  lifecycle({
+    componentDidMount() {
+      const {
+        customMiddleware,
+      } = this.props;
+
+      if (IS_AUTH_PORTAL) {
+        // only enabled if AWG_TOKEN_RENEWAL_INTERVAL is set
+        this.AWGSessionTimer = keepAliveAWGSession();
```
caravinci, line 172: Allows fence polling, for testing purposes only.

src/packages/@ncigdc/utils/ajax/index.js
```diff
@@ -1,26 +1,31 @@
 // @flow
-import { CALL_API } from 'redux-api-middleware';
+import { RSAA } from 'redux-api-middleware';
```
caravinci, line 2: Updating this dependency to the currently latest version for great good 💰

src/packages/@ncigdc/utils/auth/awg.js
```diff
@@ -0,0 +1,137 @@
+import { throttle } from 'lodash';
```
caravinci, line 1: These are all the new AWG token handlers/helpers

src/packages/@ncigdc/utils/auth/awg.js
```diff
@@ -0,0 +1,137 @@
...
+const resetSessionRenewalTimestamp = now => (
+  removeSessionRenewalTimestamp() &&
+  createSessionRenewalTimestamp(now)
+);
+
+/* this is meant to avoid duplicate queries, because having multiple Relay Environments
+ * means renewal race conditions not worth dealing with at this point. ¯\_(ツ)_/¯
+ */
+let queryCache = null;
+const handleQueryCache = value => {
```
caravinci, line 33: Not particularly happy having to do this bit, but seems to get the job done to ensure no unnecessary duplicate calls from race conditions between the coexisting Relay environments.

src/packages/@ncigdc/utils/auth/awg.js
```diff
@@ -0,0 +1,137 @@
...
+  return !localStorage.awg_sessionLastRenewed;
+};
+
+const resetSessionRenewalTimestamp = now => (
+  removeSessionRenewalTimestamp() &&
+  createSessionRenewalTimestamp(now)
+);
+
+/* this is meant to avoid duplicate queries, because having multiple Relay Environments
+ * means renewal race conditions not worth dealing with at this point. ¯\_(ツ)_/¯
```
caravinci, line 30: Redesign is nigh! 🎉

src/packages/@ncigdc/utils/consoleDebug.tsx
```diff
@@ -1,9 +1,12 @@
...
+import {
+  DEBUGGING,
+  IS_DEV,
+} from '@ncigdc/utils/constants';
 
 type TConsoleDebug = (log: string) => void;
 
 const consoleDebug: TConsoleDebug = log => {
-  if (IS_DEV) {
+  if (IS_DEV || DEBUGGING) {
```
caravinci, line 9: This will allow flag-based prod console logging, for better testing and debugging.

.travis.yml
```diff
@@ -15,7 +16,8 @@ jobs:
     # this will only run lint / tests against modified files but it is unstable
     # - echo $TRAVIS_COMMIT_RANGE | sed 's/\.\.\./../' | xargs git diff-tree --no-commit-id --name-only -r | xargs echo -n
     - npm run lint || travis_terminate 1
-    - npm test || travis_terminate 1
+    # These tests are largely out of date and the coverage is not worth the build time. commenting
+    # - npm test || travis_terminate 1
```
singergr, line 20: Would it be possible to leave the tests in, and make this change in a separate PR if needed?

caravinci: Absolutely! 👍

src/packages/@ncigdc/utils/auth/awg.js
```diff
@@ -0,0 +1,137 @@
...
+};
+
+const renewAWGToken = throttle(({
+  now,
+  origin,
+  timeSinceLastRenewal,
+}) => {
+  consoleDebug(`Querying AWG fence at ${now}, by ${origin}`);
+
+  return queryCache || fetch(urlJoin(FENCE, 'user'), {
```
philtom-ctds, line 51: `queryCache` could be `err` from the line 74, which will be truthy.  But `queryCache` could be falsey from line 63 when the fence request fails.  Is it intentional this to behave like a five-second debounce for thrown errors but not debounce when the previous request is not `ok`?

src/packages/@ncigdc/utils/ajax/index.js
philtom-ctds: If you like this, you should checkout `createAsyncThunk` from redux-toolkit.  it's beautiful.

src/packages/@ncigdc/utils/auth/awg.js
caravinci: That's right. The idea here (if you don't mind me breaking it down, just to double check my logic) is that if `queryCache` is `null`, it means a request hasn't happened in the last 5 seconds. 
If instead it has "any" value, then there's another request "going on" (not really, as I'm straight up giving the response if available...), so don't make a new call.
In other words, I'm micromanaging asynchronicity and runtime delays created by technical debt.
Part of the reason why I chose to update some of the dependencies.

How it happens:
There cases in which multiple UI components are making almost simultaneous requests, with a second or less of difference in between them, which are not getting caught by `lodash.throttle`, before the first one updates the timestamp to prevent the other requests from happening. If I extend the throttling, then those other attempts are rejected, and the whole app flow crashes, logging the user out immediately, even though the fence validates the first call.
This is as close as a sweet spot as I managed to find.

To your question, the result itself being `truthy` or `falsy` or even an error, is not a concern here, as the respective handling is happening elsewhere. It's more avoiding the duplication without collapsing the chain of actions dispatched.

Does any of this make sense? lol

src/packages/@ncigdc/utils/ajax/index.js
caravinci: Personally, not a fan of `redux-api-middleware`; but I hadn't seen `createAsyncThunk` and it looks like a thing of beauty!

src/packages/@ncigdc/utils/auth/awg.js
```diff
@@ -0,0 +1,137 @@
...
+            now,
+            origin,
+            timeSinceLastRenewal,
+          })
+        );
+  }
+  return !AWG; // do not interfere if not on AWG
+};
+
+export const clearAWGSession = timerId => {
```
singergr, line 108: I don't think anything other than the `RelaySetup` component knows the timer ID to pass it into `clearAWGSession`. If something else calls this before `RelaySetup` is unmounted, it will clear `localStorage.awg_sessionLastRenewed`, and then this won't clear the timer when `RelaySetup` unmounts. It might be better to keep the timer management in `Root.js`.

This is not a blocking issue since we shouldn't be enabling the refresh timer in any actual deployments anyway.

src/packages/@ncigdc/utils/auth/awg.js
```diff
@@ -0,0 +1,137 @@
...
+      throw handleQueryCache(err);
+    });
+}, 1000);
+
+export const checkAWGSession = (renewalType = 'manual', origin = 'unknown origin') => {
+  if (localStorage.awg_sessionLastRenewed) {
+    const nowMoment = moment.utc();
+    const now = nowMoment.format();
+    const timeSinceLastRenewal = nowMoment.diff(moment(
+      localStorage.awg_sessionLastRenewed.replace('-pending', ''), // '-pending' throttles queries
```
singergr, line 83: I don't see where `-pending` is appended. Is it intended to be appended somewhere, or is this left over from an earlier revision?

src/packages/@ncigdc/routes/Login.js
```diff
@@ -59,112 +67,128 @@ const AWGLoginButton = compose(
           return (window.location.href = '/login?error=no_fence_projects');
         }
       }
-      await dispatch(fetchUser());
-      push({ pathname: '/repository' });
+      const createdSession = await createAWGSession();
```
philtom-ctds, line 70: I am looking at this PR from a multi-tab perspective.  (when a user has multiple awg tabs open).  

If this line is executed when a user already has a tab with an active session, then the existing session will get wiped out, due to `removeSessionRenewalTimestamp()`.  But this will only matter if the user failed login from the `openAuthWindow` call above.
1. `openAuthWindow` can reject the promise with 'No cookies enabled'.  
2. `openAuthWindow` can also reject the promise if any errors are thrown inside of it. The only potential error I see is from the use of es6 code in an older browser.  (I haven't looked to see if we use babel to transpile to es5 or polyfill or similar)

Based on these two things, I'd suggest having default behavior at the end of the catch to handle any other errors from auth.   Maybe redirecting to login?

src/packages/@ncigdc/utils/auth/awg.js
```diff
@@ -0,0 +1,137 @@
...
+  queryCache = value;
+
+  setTimeout(() => {
+    queryCache = null;
+  }, 5000);
+
+  return queryCache;
+};
+
+const renewAWGToken = throttle(({
```
singergr, line 44: Could you confirm I understand this function's intended interface correctly?

1. If the renewal succeeds, `renewAWGToken` resolves to true.
2. If the HTTP request returns 403, `renewAWGToken` resolves to false, so `checkAWGSession` also resolves to false, and the upstream caller then redirects to the login page.
3. If the HTTP request doesn't get a response at all, `renewAWGToken` throws, which probably causes something bad to happen, but the situation is likely unrecoverable at that point anyway.

If so, I think 1 and 2 are fine, but to Phil's point, it'd probably be better to cache false in 3 so future calls return the cached failure. Depending on how the exception is handled, though, it might not matter.

philtom-ctds (approved):
The core logic looks good.  I have one comment from the mulit-tab perspective.  I didn't see any other obvious issues from that perspective.  You might want to give it a once over with multi-tab in mind since you are more familiar with this code than I am.  The cache/throttle code will only work per tab, but that's probably ok.  

I don't think this should block this PR from being merged.  But any multi-tab issues that might cause the user's session to get invalidated on accident should be fixed in a follow up PR while we are in QA.  Approving.

src/packages/@ncigdc/utils/auth/awg.js
singergr: The explanation makes sense to me. It seems like `_.debounce` should cover this, but it sounds like Lodash wasn't reliably grouping the calls for some reason. I'll think about it a little more and see if I'm maybe missing something, though.

singergr (commented):
I still want to check one case I'm not sure about before approving, but I had some questions in the meantime. I don't think any of these are blocking.

src/packages/@ncigdc/routes/Login.js
caravinci: Hm though provoking indeed!

Will have to study it a bit more in depth, hopefully over the upcoming RC tag.
One thing I can clear already is that all our code is transpiled/polyfilled correctly on build, so we needn't worry too much about users in older browsers (at least not on that specific subject.)

src/packages/@ncigdc/utils/auth/awg.js
caravinci: No, you're right. Somehow lodash is not grouping correctly all the calls made from different callers.
My best guess is it is some sort of transpilation issue, in which the throttled function is getting different identifiers in different contexts, and Lodash is not associating them as the same.

Could not test this theory locally (pre-transpilation) since we are not able/allowed to run auth on local machine.

src/packages/@ncigdc/utils/auth/awg.js
caravinci: Hm! I hadn't understood 3! You're both 100% right.
For documentation's sake, if the request fails, I'm passing the error response from the server (which could be anything, e.g. an object, a string or an Error entity, all of which are truthy).
Setting the queryCache to any of those will be true, and result in allowing the subsequent API call which this token check should disallow. That other request in turn will reject, and thus handling the logout.

Though that manages the result as per spec (user is logged out of the UI forcefully), this check is not preventing the other API call.

singergr (commented):
This is still missing the logic to renew the session if the `awg_access_token` cookie expired. It works in the test configuration that has been used so far, but with the production configuration, it will still be possible for users' logins to expire even if they have been active. I'll explain more offline, but we would either want to add that to this PR or plan to add it in a follow-up.

src/packages/@ncigdc/utils/auth/awg.js
caravinci: It was leftover from another attempt at the queryCache.
Left it here thinking of amending this logic to not only debounce by holding a cache with the response, but by also setting this differentiator in the cookie (which would also potentially help with the 'multi-tab' case @philtom-ctds pointed out. Figured I could continue studying this approach in a subsequent RC, as I wasn't able to make it work here)

src/packages/@ncigdc/utils/auth/awg.js
caravinci: There's no timer in normal runtime. the timer is only even created if we're in testing/debugging mode.
This is functionality I've added to help me work. Not functional in prod (unless enabled)

singergr (approved):
🤞

src/Root.js
```diff
@@ -162,58 +88,169 @@ const HasUser = connect(state => state.auth)(props => {
...
+      const hash =
+        parse(search).hash ||
+        md5(
+          [req.relayReqObj._printedQuery.text, JSON.stringify(req.relayReqObj._printedQuery.variables)].join(':'),
+        );
+
+      req.url = `${url}?hash=${hash}`;
+
+      const parsedBody = JSON.parse(req.body);
+      req.body = JSON.stringify(parsedBody);
```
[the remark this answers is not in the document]
caravinci: TBH, it now seems superfluous to me to parse the body just to stringify it again.
Looking at theses lines' original commit, looks like cruft added [here](https://github.com/NCI-GDC/portal-ui/commit/dd77f29c63ef4b0959ebfe7d8b673fa3e025ef57#diff-0d38dc387dd6ab8fac2828286d346568R74-R75), that I'm just relocating. 
It's harmless, but don't want to call it "redundant" until i'm sure.
Will make sure to check before removing it in a further commit. ✅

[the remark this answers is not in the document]
singergr: Sounds good. Assuming the request body is always JSON (which it should be for Relay), I agree it seems harmless for now.

caravinci:
For follow up, we've discussed over Slack that we would address the `awg_access_token` cookie expiration through setting the renewal interval to 20 minutes, to match the timeout settings prod. 
The implementation used 5 minutes, to match the testing environment. 
That mismatch would've allowed the unhandled scenario. Will amend in a further commit if necessary, but this was now good enough for the limelight. :shipit:

PR #2927, opened 2020-09-25
[PRTL-3236] remove malfunctioning dispatch thens

caravinci:

- [ ] `prod`
- [ ] `dev-oicr`
- [x] `qa-purple`

## Description of Changes

TL;DR: a few Redux action dispatchers were being incorrectly used as `thenables`, even though they were all dispatching a thunk without returning a promise.

Checked all dispatchers throughout the app, to ensure their validity.
Found only 4 offending dispatchers, all of which used the same single action thunk related to analyses.
...

Explanation of the bug: In order for a Redux dispatch to be made "thenable", the action creator MUST return a promise (and may be done with observables?); which was never the case in the codebase, yet somehow worked until now. ¯\(ツ)/¯

Such dispatchers stopped working as a result of updating some dependencies when working on AWG token renewal (needed to correct a race condition that blocked implementation).

Redux-api-middleware, in an outdated beta version, while fixing a bug they introduced now enforces the correct dispatch behaviour (i.e. it was erroneously returning promises, [as seen here](https://github.com/agraboso/redux-api-middleware/pull/167/files)).
That bug not being introduced anymore by the middleware, our faulty segments of code was "held accountable".

### DIFF
--- a/src/packages/@ncigdc/components/CreateOrOpenAnalysis.js
+++ b/src/packages/@ncigdc/components/CreateOrOpenAnalysis.js
@@ -51,16 +51,16 @@ const CreateOrOpenAnalysis = ({
             type,
             created,
           }),
-        ).then(() => {
-          window.open(
-            `analysis?${queryString.stringify({
-              analysisTableTab: 'result',
-              analysisId: id,
-            })}`,
-          );
-        });
+        );
+
+        window.open(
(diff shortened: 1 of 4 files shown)


anmaxvl (approved):
thanks for explanation. lgtm.

PR #2928, opened 2020-09-25
[PRTL-3237] rename AWG renewal variable for CRA

caravinci:

- [ ] `prod`
- [x] `dev-oicr`
- [x] `qa-purple`

## Description of Changes
Renames the environment variable, according to the Create React App env. validation requirement.


src/packages/@ncigdc/utils/constants.js
```diff
@@ -816,7 +816,7 @@ export const DEV_USER_CA = [ // controlled access mock
 
 export const AWG_TOKEN_EXPIRY = (
   localStorage.AWG_TOKEN_EXPIRY ||
```
singergr, line 818: I'm not sure if you might want the local storage variable to have the same prefix for consistency, but it shouldn't affect the deployment at all, so I don't see an issue either way.

singergr (approved):
Haven't tested, but it looks like it should work.

`build.sh` still unsets the old variable, but unless someone defined `REACT_APP_AWG_LOGIN_EXPIRY` before running `build.sh`, that shouldn't matter.

src/packages/@ncigdc/utils/constants.js
caravinci, line 818:
```suggestion
  localStorage.REACT_APP_AWG_TOKEN_EXPIRY ||
```

caravinci: Thought I'd keep it without prefix for ease of use while debugging, but you're right... it should be consistent.

PR #2929, opened 2020-09-30
[PRTL-3235] mean centering of colours in GE heatmap

caravinci:

- [ ] `prod`
- [ ] `dev-oicr`
- [x] `qa-purple`

## Description of Changes
Recalculates the colour ranges around 0 (rather than finding the middle value between min and max), to match our median centring algorithm.


samrichca (approved):
LGTM

PR #2936, opened 2020-10-01
[PRTL-3244] outer dendrogram line enhancements

samrichca:
## Ticket Number

PRTL-3244

- [ ] `prod`
- [ ] `dev-oicr`
- [x] `qa-purple` for gene ex data

## Description of Changes

- determine outer dendrogram line more reliably
- remove click events and cursor
- change text of tooltips
- handle situations like clicking on inner dendrogram to select it -> then clicking on outer dendrogram to de-select the first one
- reposition tooltips for row dendrograms

### DIFF
--- a/src/packages/@ncigdc/modern_components/GeneExpression/inchlib/index.js
+++ b/src/packages/@ncigdc/modern_components/GeneExpression/inchlib/index.js
@@ -1720,11 +1720,18 @@ import { getLowerAgeYears } from '@ncigdc/utils/ageDisplay';
     return nodes2columns;
   };
 
+  InCHlib.prototype._check_top_node = function (id) {
+    return id.includes('_top_node');
+  }
+
   InCHlib.prototype._bind_dendrogram_hover_events = function (layer) {
     const self = this;
 
     layer.on('mouseover', function (evt) {
-      self._cursor_mouseover();
+      const is_top_node = self._check_top_node(evt.target.attrs.id);
+      if (!is_top_node) {
+        self._cursor_mouseover();
+      }
       self._dendrogram_layers_mouseover(this, evt);
     });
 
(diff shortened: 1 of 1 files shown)


src/packages/@ncigdc/modern_components/GeneExpression/inchlib/index.js
```diff
@@ -2642,38 +2650,45 @@ import { getLowerAgeYears } from '@ncigdc/utils/ageDisplay';
     });
   };
 
-  InCHlib.prototype._highlight_cluster = function (path_id) {
+  InCHlib.prototype._highlight_cluster = function (path_id, id) {
```
caravinci, line 2653: What's the difference between those two ids?

src/packages/@ncigdc/modern_components/GeneExpression/inchlib/index.js
```diff
@@ -1589,8 +1589,7 @@ import { getLowerAgeYears } from '@ncigdc/utils/ageDisplay';
       const y1 = self._get_y1(node_neighbourhood, current_left_count, current_right_count);
       let y2 = self._get_y2(node_neighbourhood, current_left_count, current_right_count);
       let x1 = self._hack_round(self.distance - self.distance_step * node.distance);
-      x1 = (x1 == 0) ? 2 : x1;
-
+      x1 = (x1 === 0) ? 2 : x1;
```
caravinci, line 1592: 🙌

src/packages/@ncigdc/modern_components/GeneExpression/inchlib/index.js
samrichca: the `path_id` is used to construct the `id` and doesn't contain all of the same information - in this case it doesn't contain `_top_node` for the top dendrogram node

caravinci: meaning, the `path_id` is sort of the original, and then the id would be the one actually used in the DOM.
is that correct?

samrichca: something like that. i figured it was safer to customize the constructed, final ID than its constituent parts.

caravinci (approved):
Squasharoooo 🦘

PR #2948, opened 2020-10-19
[PRTL-3262] correct counts on projects graph

samrichca:
## Ticket Number

PRTL-3262

- [x] `prod`
- [ ] `dev-oicr`
- [ ] `qa` (which version?)

## Description of Changes

correct counts on projects graph


caravinci (approved):
declarative af! (as functional)

PR #2962, opened 2020-12-15
Chore/bump version

scintillating7:

- [ ] `prod`
- [ ] `dev-oicr`
- [x] `qa` purple

## Description of Changes

We need to exclude changes since wocket (aka gene expression), but still bump the base images and bump the version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.25.1",
+  "version": "1.27.0-rc.2",
   "repository": "NCI-GDC/portal-ui",
   "name": "portal-ui",
   "author": {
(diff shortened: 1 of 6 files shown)


a-collet (approved):

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.25.1",
+  "version": "1.27.0-rc.2",
```
philtom-ctds, line 2: We don't need to update this version for every release candidate.  We should be able to update it to 1.27.0 now.

philtom-ctds (approved):

PR #2966, opened 2021-01-19
Add registry arg in docker builds for portal

jawadqur:
## Ticket Number

GDCDEVOPS-572

## Description of Changes
Add registry argument in docker builds for portal. These dockerfiles are only used in Jenkins and not travis.

### DIFF
--- a/awgportal.Dockerfile
+++ b/awgportal.Dockerfile
@@ -1,3 +1,4 @@
+ARG registry=docker.osdc.io
 FROM node:13 as builder
 
 WORKDIR /portal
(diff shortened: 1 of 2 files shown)


singergr (approved):
I got totally confused by the fact that the Travis build uses a different Dockerfile. lol.

jawadqur:
> I got totally confused by the fact that the Travis build uses a different Dockerfile. lol.

Yeah we should change Travis behavior to not do the node builds outside of docker build and leverage the dockerfiles

kulgan (approved):

PR #2968, opened 2021-02-20
DEV-524 support range facet for doubles

philtom-ctds:
In DR 27, we updated some of mapping property types to double. This
was needed to perform aggregation without bucketing into integers.
As a result, custom facets broke because they did not support doubles.

### DIFF
--- a/src/packages/@ncigdc/components/__tests__/FacetWrapper.tests.js
+++ b/src/packages/@ncigdc/components/__tests__/FacetWrapper.tests.js
@@ -1,0 +1,17 @@
+import { getFacetType } from '@ncigdc/components/FacetWrapper';
+
+describe('getFacetType', () => {
+  describe('returns "range"', () => {
+    it('for long types', () => {
+      expect(getFacetType({ type: 'long' })).toBe('range');
+    });
+
+    it('for float types', () => {
+      expect(getFacetType({ type: 'float' })).toBe('range');
+    });
+
+    it('for double types', () => {
+      expect(getFacetType({ type: 'double' })).toBe('range');
+    });
+  });
+});
--- a/src/packages/@ncigdc/containers/FileAggregations.js
+++ b/src/packages/@ncigdc/containers/FileAggregations.js
@@ -147,7 +147,7 @@ export type TProps = {
     doc_type: String,
     field: String,
     full: String,
-    type: 'id' | 'string' | 'long',
+    type: 'id' | 'string' | 'long' | 'float' | 'double',
   }>,
   handleSelectFacet: Function,
   handleResetFacets: Function,
--- a/package-lock.json
+++ b/package-lock.json
@@ -4555,19 +4555,30 @@
           }
         },
         "chokidar": {
-          "version": "3.4.2",
-          "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-3.4.2.tgz",
-          "integrity": "sha512-IZHaDeBeI+sZJRX7lGcXsdzgvZqKv6sECqsbErJA4mHWfpRrD8B97kSFN4cQz6nGBGiuFia1MKR4d6c1o8Cv7A==",
+          "version": "3.5.1",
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -145,6 +145,6 @@
     "fsevents": "^2.0.7"
   },
   "engines": {
-    "node": ">=13.2.0"
+    "node": ">=12.0.0"
```
scintillating7, line 148: We're now accepting lower versions of node? 🤔

src/packages/@ncigdc/containers/FileAggregations.js
```diff
@@ -147,7 +147,7 @@ export type TProps = {
     doc_type: String,
     field: String,
     full: String,
-    type: 'id' | 'string' | 'long',
+    type: 'id' | 'string' | 'long' | 'float' | 'double',
```
scintillating7, line 150: Interesting that `float` wasn't in there before

package.json
jiakf: not part of this PR, but looked up "why-did-you-update" dependency because I was curious what it did. Looks like its deprecated though: if there's a ticket for updating dependencies for portal-ui, could probably add this

philtom-ctds: i can revert this. however, it's a bad practice to depend on odd versions of node.  those are basically development versions. the even version numbers are the stable and lts versions.  going down to 12 was easier than going up to 14. i'll create a separate ticket for updating the node version.  there's other things to be updated, such as docker.

src/packages/@ncigdc/containers/FileAggregations.js
philtom-ctds: These are typings for flow. They don't actually change how the js executes.  But since these typing existed, I updated them to be more accurate.  fwiw, flow lost to typescript in the "typed javascript" war.

src/packages/@ncigdc/components/__tests__/FacetWrapper.tests.js
```diff
@@ -0,0 +1,17 @@
+import { getFacetType } from '@ncigdc/components/FacetWrapper';
+
+describe('getFacetType', () => {
+  describe('returns "range"', () => {
```
singergr, line 4: We could test returning other facet types; otoh we had no tests at all before.

singergr (commented):
Seems okay to me, but two quick questions:

1. Do we still want to update package-lock?
2. There is a similar function in `src/packages/@ncigdc/utils/getFacetType.js`. Am I right that the other function is unused?

kulgan (approved):
lgtm

BenX00 (approved):
lgtm

src/packages/@ncigdc/components/__tests__/FacetWrapper.tests.js
philtom-ctds: I had contemplated adding more tests, but that code has some oddities to it. Some of it appears to be outdated and the order of the if blocks appear to matter.  This makes the intent of this function a little ambiguous.  Due to that, I didn't want to write tests for the existing behavior that I couldn't verify as correct and limited the tests to the scope of the bug.

philtom-ctds:
> 1. Do we still want to update package-lock?
Yes. These updates only affect local development and were independent from the previous node engine version change that I reverted earlier.

The `package.json` preinstall script:
```
    "preinstall": "npx npm-force-resolutions",
```
will force installation of the newest version that matches these definitions from `package.json`:
```
  "resolutions": {
    "@types/relay-runtime": "1.3.10",
    "eslint": "^6.8.0",
    "chokidar": "^3.0.0"
  },
```
and update the `package-lock.json`.

> 2. There is a similar function in `src/packages/@ncigdc/utils/getFacetType.js`. Am I right that the other function is unused?

You are correct.  That function is not used. If it were used, we'd see the `getFacetType` module `import`ed or `require`d somewhere in the codebase. (of course, this is javascript. even the imports can be dynamically generated at runtime. so, there's always a chance.)

PR #2979, opened 2021-04-21
Removed npm-force-resolutions  from preinstall

bilalbaqar:

- [ ] `prod`
- [ ] `dev-oicr`
- [x] `qa` (which version?)

## Description of Changes

We use npm-force-resolutions to modify package-lock.json in npm preinstall. This step `npx npm-force-resolutions` has been failing when run from behind our proxy due to (what i think is) this [hard timeout.](https://github.com/rogeriochaves/npm-force-resolutions/blob/master/src/npm_force_resolutions/core.cljs#L70) I believe this is true because the command fails in 8-9secs. Also I when i used an older version of npm-force-resolutions which doesnt have that timeout, it passes. 

Filed issue https://github.com/rogeriochaves/npm-force-resolutions/issues/33 and PR https://github.com/rogeriochaves/npm-force-resolutions/pull/34/files  in upstream repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "homepage": "https://portal.gdc.cancer.gov/",
   "scripts": {
-    "preinstall": "npx npm-force-resolutions",
+    "preinstall": "npx npm-force-resolutions@0.0.3",
     "start": "concurrently --kill-others 'npm run start-ui' 'npm run relay -- --watch'",
     "start-ui": "BROWSER=none react-app-rewired start --scripts-version react-scripts-ts",
     "build": "npm run relay && react-app-rewired build --scripts-version react-scripts-ts",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "homepage": "https://portal.gdc.cancer.gov/",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "homepage": "https://portal.gdc.cancer.gov/",
   "scripts": {
-    "preinstall": "npx npm-force-resolutions",
+    "preinstall": "npx npm-force-resolutions@0.0.3",
```
philtom-ctds, line 11: Could you try one more thing?  Renamed `preinstall` to `force-resolve`.  This should remove `npm-force-resolutions` from the install stages, which should also solve this CI issue we're seeing.

My rationale is that builds should be deterministic and we should not force an update of dependencies at CI/CD build time.  Instead, this should be an intentional action done by a human, which can then be tested and reviewed.

bilalbaqar: Yeah that does make sense. So are we okay with npm-force-resolutions not running at build time? I am guessing you can run it manually to force changes when needed?

scintillating7 (approved):

philtom-ctds (approved):

PR #3014, opened 2022-06-23
PEAR-382 bumping version

philtom-ctds:


BenX00 (approved):

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.28.0",
+  "version": "1.29.0",
```
mpsolano, line 2: I think this should be:

```suggestion
  "version": "1.30.0",
```

jiakf (approved):

mpsolano (changes requested):

package.json
philtom-ctds: updated

mpsolano (approved):
Looks good!

PR #3043, opened 2023-06-15
DEV-1822 try to see if this fixes the ID problem

XukunXiangUChicago:
with the graphene v1.4.1, it is now asking for `id: ID!` for the node field.


mpsolano (approved):
Looks good!

stilesj-uchicago (approved):

